PR #3, opened 2021-05-16
New Question: How has stopping pooled rideshare trips affected communities?

vingkan:
## Overview of Changes

- For new question:
  - [x] Added data about pooled rideshare trips and average cost to `rideshare` table
  - [x] Added `PooledTripMetrics` class to API with unit tests
  - [x] Added new question page at `questions/pooled-trips` using bar chart and sortable table
- For supporting code:
  - [x] Added `Table` component with sorting
  - [x] Fixed unit tests and build tests to run on pull requests
  - [x] Added method to `CommunityMetrics` to merge multiple metrics by area

## Question Preview

localhost_8001_transithealth_questions_pooled-trips


Andrew-Tolentino:
👀

vingkan:
Merging for now to get an example question on the site and to avoid this PR becoming even more bloated.

One big piece of tech debt to address: the GET /question/pooled_trips endpoint is very slow. Maybe running five aggregation queries on a 3.7M row table was a bad idea? Need to think some more about how to make this a better experience.

Some ideas:
- Generate this response in the offline pipeline and serve the data directly without running queries online
- Combine the metrics into one query so it can be done in one pass rather than five separate passes
- Change the rideshare table to be on a weekly granularity rather than a daily granularity
- Parallelize the metric computations
- Look into a caching strategy for the API (this won't solve the first load wait time, but it might make the developer experience working on this feature more bareable)
- Have a great loading screen interaction so the user *wants* to wait lol
- Split the data into separate requests and gracefully load the content for the question page
- Store the data for the first bar chart in the client (or at least as a public resource file that the CDN will cache) and then gracefully load the table from the API

PR #4, opened 2021-05-21
Add median income metric

Andrew-Tolentino:
## Technical Description
* Added median income metric from Chicago Atlas Data APIs in Offline Pipeline
* Created backend API to query median income from income table
* Added median income metric to ScatterPlot option on the front end


api/metrics/community_test.py
```diff
@@ -44,6 +44,34 @@ def test_population():
...
+            "value": 10000
+        }
+    ]
+    con, cur = create_test_db(
+        scripts=[
+            "./pipeline/load/income.sql"
+        ],
+        tables={
+            "income": income_table
+        }
```
vingkan, line 74: What happened to the test assertions? When I looked yesterday I saw you had them in this method, maybe they got removed while fixing merge conflicts or something.

pipeline/Makefile
```diff
@@ -40,6 +40,7 @@ database.db: data/loaded/all_tables.txt
 data/loaded/all_tables.txt:
 	make data/loaded/community_area.txt
 	make data/loaded/population.txt
+	make data/loaded/income.txt
```
vingkan, line 43: Makefile change looks great, good job!

pipeline/load/income.sql
```diff
@@ -0,0 +1,10 @@
+DROP TABLE IF EXISTS income;
+CREATE TABLE income (
```
vingkan, line 2: Want to add a primary key for this table?

pipeline/tests/database_test.py
```diff
@@ -21,6 +21,7 @@
 def test_table_counts():
     test.assert_table_count("community_area", 77)
     test.assert_table_count("population", 770)
+    test.assert_table_count("income", 770)
```
vingkan, line 24: Thanks for adding these, will help make sure future changes to the database have the right tables.

vingkan (approved):
Looks great! This is a great example of the changes needed to add a dataset and metrics.

One actionable comment about how the test assertions from your API test seem to have disappeared, but after you add those and they pass, you are free to merge.

Approving this PR now so you don't have to wait for me after you fix the test.

api/metrics/community_test.py
Andrew-Tolentino: Oh darn, will look into it

Andrew-Tolentino: As of right now, would you be okay with having identical metric tests in this file for the income and population tests? Made it identical since the methods are almost the same and that it seems like these tests are unit tests since they don't actual check into the database but an in-memory one.

vingkan: Yeah it's fine for them to be almost identical!

And yes, you're right that it's testing an in-memory database using the test data you declared. Just helps us test the metric logic on a controlled (and made up) dataset.

PR #8, opened 2021-06-03
Update Jupyter Notebook template to run 3rd cell

Andrew-Tolentino:
Made a quick fix to the 3rd cell in default notebook
* Updated table reference of `rideshares` to `rideshare`
* Pass db connection into `RideshareMetrics` object instead of cursor


vingkan (approved):
Thanks for catching and fixing this!

PR #29, opened 2021-07-12
Rearrange instructions for notebook setup.

vingkan:
Sorry! In the agenda for the first mentor meeting, I didn't include the instructions for setting up the interns' Jupyter notebook folders and submitting a pull request for that. Thanks @rbrooks6 for catching that.

The notebook setup instructions were buried in the old setup docs that we no longer use since we now have Cloud9. I made a new page that walks people through setting up their notebooks folder.

Now, you can also start your Jupyter notebook server with the `run notebooks 9` command on Cloud9.

### DIFF
--- a/api/dev.sh
+++ b/api/dev.sh
@@ -17,4 +17,4 @@ export FLASK_DEBUG=1
 export FLASK_ENV=development
 export FLASK_RUN_HOST=0.0.0.0
 export FLASK_RUN_PORT=5000
-flask run || echo "Failed to start Flask. Is your virtual environment activated?"
+flask run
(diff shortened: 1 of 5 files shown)


vingkan:
It would be nice to not have to copy and paste the token on Cloud9, but I couldn't figure out how to automate that.

My idea was:

```
# Run the notebook server in the background
jupyter notebook &
# List the running servers and their tokens, somehow get the token from that output
TOKEN=$(jupyter notebook list | grep ???)
# Display the link to automatically authenticate with token
echo "http://$INSTANCE_IP:8080?token=$TOKEN"
```

But I didn't manage the background process correctly, so the content getting printed to the terminal was out of order and the background process didn't get stopped when I pressed Ctrl+C, causing the next usage to start on port 8081. I had to manually get the pids and stop the notebook servers.

PR #30, opened 2021-07-12
Implement Belonging Extract Step

fabrego524:
Modified makefile and implemented belonging.py using income.py as basis

### DIFF
--- a/pipeline/extract/belonging.py
+++ b/pipeline/extract/belonging.py
@@ -1,0 +1,57 @@
+import sys
+sys.path.append("./")
+
+import argparse
+import json
+import os
+import pandas as pd
+import requests
+from time import sleep
+from timeit import default_timer as timer
+from tqdm import tqdm
+from utils.data import HEALTH_ATLAS_VALUE_COLS
+
+
+cli = argparse.ArgumentParser(description="Download belonging data for each community area.")
+cli.add_argument("--coverage_file", help="File path to read coverages from.")
+cli.add_argument("--output_file", help="File path to write results to.")
+args = cli.parse_args()
+
+start = timer()
+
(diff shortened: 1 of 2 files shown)


pipeline/extract/belonging.py
```diff
@@ -0,0 +1,57 @@
...
+
+with open(args.coverage_file) as coverage_file:
+    coverage_json = json.loads(coverage_file.read())
+    df_coverage = pd.DataFrame(coverage_json["coverages"]["neighborhood"])
+    n_periods = len(df_coverage["period"].unique())
+    n_populations = len(df_coverage["population"].unique())
+    print(f"Population data covers {n_periods} periods and {n_populations} populations.")
+
+# Get periods only for the full population (empty string)
+periods_full_population = ["2016-2018","2015-2017"]
```
vingkan, line 32: When we were pairing, I wanted to explicitly write out the year periods. But that means we are not actually getting them from the coverage file.

Let's change line 32 back to:

```python
# Get periods only for the full population (empty string)
periods_full_population = df_coverage[df_coverage["population"] == ""]["period"].values
```

This will have the same result, since `["2016-2018","2015-2017"]` are the only periods covered for this topic. However, now we will get them from the coverage file instead of hardcoding it. That will let us add other periods and population segments later, while only requesting those that are covered.

After changing this, to check if the extract step still works, try out these commands:

```
# activate virtual environment and go to pipeline (if you haven't already)
source .venv/bin/activate
cd pipeline
# rerun the extract step
rm data/extracted/belonging.csv
make data/extracted/belonging.csv
```

Then open the .csv file and check if the data is as you expect.

vingkan (changes requested):

vingkan (approved):
Nice work, Fabian!

PR #33, opened 2021-07-13
Implement Belonging Transform Operations

fabrego524:
Modified makefile and implemented belonging.py in transform using income.py in transform as a basis. Added clarification to documentation in belonging.py regarding the split of the geoid.


vingkan (approved):
Excellent, looks good to merge.

PR #34, opened 2021-07-13
Transform disabilities

MariaGarciaCS:
Created disabilities.py using income.py as a template and modified Makefile to output transformed disabilities data.


vingkan (approved):
Looks good! Merge it when you are ready.

PR #37, opened 2021-07-13
Implement Belonging Load Operation

fabrego524:
Implemented using a similar schema, modified makefile with belonging.txt placed using existing convention. Results tested and compared using SQLite in Cloud9, data is properly displayed and matches belonging.csv data.

### DIFF
--- a/pipeline/Makefile
+++ b/pipeline/Makefile
@@ -40,6 +40,7 @@ database.db: data/loaded/all_tables.txt
 data/loaded/all_tables.txt:
 	mkdir -p data/loaded
 	make data/loaded/community_area.txt
+	make data/loaded/belonging.txt
 	make data/loaded/population.txt
 	make data/loaded/income.txt
 	make data/loaded/covid_spread.txt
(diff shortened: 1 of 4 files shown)


pipeline/load/belonging.sql
```diff
@@ -0,0 +1,11 @@
+DROP TABLE IF EXISTS belonging;
+CREATE TABLE belonging (
+    area_number INTEGER,
+    segment TEXT,
+    value REAL,
+    std_error REAL,
+    period_start_year INTEGER,
+    period_end_year INTEGER,
+    period TEXT,
+    PRIMARY KEY(area_number)
```
vingkan, line 10: This works! However, the primary key should be the column(s) that uniquely identify rows in the table.

In this case, I would suggest:

```
PRIMARY KEY (area_number, period_end_year)
```

I left `segment` out of the primary key because right now we only have one segment: the entire population.

pipeline/Makefile
```diff
@@ -81,6 +82,13 @@ data/loaded/rideshare.txt: data/transformed/rideshare.csv
 		--create_sql="load/rideshare.sql" \
 		--table_name="rideshare"
 
+data/loaded/belonging.txt: data/transformed/belonging.csv
```
vingkan, line 85: Excellent, glad the load worked with no errors.

Can you please add unit tests for the new `belonging` table to the `pipeline/tests/database_test.py` file?

https://github.com/scarletstudio/transithealth/blob/main/pipeline/tests/database_test.py

Check out the tests for `income` as an example. Let's test:

- Number of rows
- Distinct values for period
- Distinct values for population

This will help others make sure they have the correct version of the database.

You can `run tests` to run all tests in the project.

vingkan (changes requested):
Great work! I requested two changes, then we can merge.

vingkan (approved):
LGTM! Just remove the `-v` from the pytest command then you are good to merge.

PR #40, opened 2021-07-13
Run command fixes

vingkan:
- Added `-vvl` flags to `run tests` so developers can see the local variables and find out which values cause the test to fail
- Fix GitHub actions config to run unit tests if any Python file changes
- Added an `else` case with message if a developer types a run command that does not have a shortcut
- Added a message to `run api 9` to print the URL with the instance IP where the API is running


fabrego524 (approved):
All looks good, the new test tag works now. Should be good to merge

PR #42, opened 2021-07-13
Run GitHub actions when database changes.

vingkan:


vingkan:
I noticed that Asude's PR changed the compressed DB files but the tests didn't run on GitHub actions, this should make tests run. Merging without approval so that Asude doesn't have to pull and merge twice.

PR #43, opened 2021-07-14
Implementation of Belonging Metric

fabrego524:
Added the implementation of the Belonging Metric into the Community Area metrics, using existing code structure as a basis and returning the community area number and the rate of belonging. Also wrote some basic unit tests.

### DIFF
--- a/api/metrics/community.py
+++ b/api/metrics/community.py
@@ -149,3 +149,20 @@ class CommunityMetrics:
         cur.execute(query)
         rows = rows_to_dicts(cur, cur.fetchall())
         return rows
+
+    def belonging(self, year, segment):
+        """
+        Returns a value between 0 and 1, the rate of residents who feel a sense of belonging for each community area
+        """
+        query = """
+        SELECT
+            area_number,
+            CAST(value AS REAL) AS value
+        FROM belonging
+        WHERE period_end_year == {year}
+        AND segment == "{segment}"
+        """.format(year=year, segment=segment)
+        cur = self.con.cursor()
+        cur.execute(query)
+        rows = rows_to_dicts(cur, cur.fetchall())
+        return rows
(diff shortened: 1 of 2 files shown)


fabrego524:
As a side note, I'm not exactly sure if
`round((CAST(value AS REAL))/100, 3) AS value`
is entirely necessary. The issue calls for the value to be returned as a value between 0 and 1. However, I've ran into an issue where values such as `35.2/100` end up with well over 10 decimal places. This is why I added the rounding to the SQL query, since I was unsure if this was some sort of division quirk with the math. Testing with SQLite on the database made it difficult to reproduce, but it is reproducible with the values in the unit test.

api/metrics/community.py
```diff
@@ -149,3 +149,20 @@ def covid_spread_sum_by_area(self, column_name):
...
         return rows
+
+    def belonging(self, year, segment):
+        """
+        Returns a value between 0 and 1, the rate of residents who feel a sense of belonging for each community area
+        """
+        query = """
+        SELECT
+            area_number,
+            round((CAST(value AS REAL))/100, 3) AS value
```
vingkan, line 160: We don't have to round the result on the backend, we can do that on the frontend.

To make your test work, instead of writing out the float, you could write the actual division:

```python
{ "area_number": 1, "value": 45.6 / 100 },
```

Let me know if that works!

vingkan (changes requested):
Good work! Change it to not round the result and if the updated tests work you can merge.

api/metrics/community.py
fabrego524: This worked perfectly for the unit test. I also removed the casting to a REAL since table schema already has it defined as such.

vingkan (approved):
LGTM!

PR #44, opened 2021-07-14
Integrate Belonging Metric with Endpoint and Config

fabrego524:
Added the Belonging Metric into the list of endpoints using existing conventions, one for 2017 and one for 2018. Also added the metrics into metrics.js with names and units.


vingkan (commented):
Looks good! We chatted offline and found a bug in the community explorer component that incorrectly handles opacity when values are missing. Turns out at least community area 47 doesn't have a value for the belonging survey. Glad that your work helped us catch this bug!

I will send a PR for that fix, you can merge that into your branch, then we can merge this PR.

vingkan (approved):
Looks good! Let's launch your first metric on the website!!

PR #45, opened 2021-07-14
Fix community explorer to correctly show opacity when values are missing.

vingkan:
Fabian's PR #44 exposed this bug in the community explorer component. If some community areas are missing data for the selected y-axis metric, then the opacity will be computed incorrectly, causing the Chicago map to show the same shade for all areas.

This PR fixes that bug:

- Filters out missing data when computing the max value
- Replaces missing data with zero when computing opacity

Here is a screenshot showing the fix applied to Fabian's branch:

Screen Shot 2021-07-14 at 3 40 13 PM

Burnside is the area 47, which is missing the belonging data. It shows up as zero opacity on the map, while the other areas show the variation.

### DIFF
--- a/app/components/CommunityScatterExplorer.js
+++ b/app/components/CommunityScatterExplorer.js
@@ -39,9 +39,13 @@ async function getScatterMetrics(metricX, metricY) {
 }
 
 function getGeoMapMetric(metrics, selectedMetric, minAlpha=0.05) {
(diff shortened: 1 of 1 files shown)


fabrego524 (approved):
This should hopefully avoid any confusion about this type of situation in the future. I went ahead and double-checked the source of this particular data, and they indicate there is no data for that particular area. I'm glad we were able to solve it. LGTM

PR #48, opened 2021-07-15
Reintroduce covid_spread.sql into pipeline

fabrego524:
Added back `covid_spread.sql` and reran the pipeline. The database is created completely, and all tests pass.

image


vingkan (approved):
Thank you for doing this! Great catch and fast fix.

PR #50, opened 2021-07-16
Offline Belonging Timeline

fabrego524:
Began backend implementation for a timeline metric for the Belonging Dataset. Modified both Belonging Extract and Transform, as well as unit tests and table schema. Some modifications of changing lists to sets

### DIFF
--- a/pipeline/Makefile
+++ b/pipeline/Makefile
@@ -205,16 +205,16 @@ data/extracted/daily_rideshare.csv:
 		--soql_file="extract/daily_rideshare.sql" \
 		--output_file="data/extracted/daily_rideshare.csv"
 	
-data/extracted/belonging.csv: data/extracted/belonging_coverage.csv
+data/extracted/belonging.csv: data/extracted/belonging_coverage.json
 	python3 extract/belonging.py \
-		--coverage_file="data/extracted/belonging_coverage.csv" \
+		--coverage_file="data/extracted/belonging_coverage.json" \
 		--output_file="data/extracted/belonging.csv"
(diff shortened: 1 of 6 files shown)


vingkan (approved):
Great work, Fabian! This is a solid example of adding new coverages for a dataset in the offline pipeline.

@Andrew-Tolentino wanted you to see this, check out how we automatically pulled the request parameters from the coverages, for both community area and city-wide data.

Because the belonging table now has both community area and city-wide data, we also had to adjust the transform code and the table schema.

Andrew-Tolentino:
> Great work, Fabian! This is a solid example of adding new coverages for a dataset in the offline pipeline. This will enable timeline metrics for belonging data, not just community area metrics.
> 
> @Andrew-Tolentino wanted you to see this, check out how we automatically pulled the request parameters from the coverages, for both community area and city-wide data.
> 
> Because the belonging table now has both community area and city-wide data, we also had to adjust the transform code and the table schema.

Oh wow this is really awesome! Way to go on configuring the request params! Out of curiosity, due to this additional layer data, is that why we needed to create a .json file instead of the .csv file?

Also, nice job on the tests too!

vingkan:
@Andrew-Tolentino On the .json vs .csv - that was actually a mistake I made when I first did the population data. The coverage response is a JSON file, we have been reading it as JSON in the extract step, but we have been incorrectly writing it in the download step as a .csv when it is actually .json. Feel free to fix it on your datasets and make the coverage file .json whenever you and your group have a chance!

PR #56, opened 2021-07-16
Fixed axis left/right bugs on timeline.

vingkan:
The timeline metric selectors show a different axis than the axis that metric is actually displayed on.

Here is a screencast (2:30) explaining one the bug and the fix:

https://www.loom.com/share/c307ef9cc09e461faa75ee5e23c36c07


MariaGarciaCS (approved):
This looks good to me! I checkout your branch, ran the application and tested a couple different scenarios, the axis changed accordingly every time. Great work fixing the left and right axis!

PR #59, opened 2021-07-16
Implement Timeline Metric for Belonging

fabrego524:
Includes initial work from PR #50 as well as two new files: `annually.py` and `annually_test.py`. These new metrics were also added to the proper endpoint and config. So far they're limited only to Race/Ethnicity populations but can be expanded to other populations of interest if needed.

### DIFF
--- a/api/metrics/annually.py
+++ b/api/metrics/annually.py
@@ -1,0 +1,29 @@
+from api.utils.database import rows_to_dicts
+
+
+class AnnuallyMetrics:
+    """
+    Metrics for annually timeline view.
+    """
+
+    def __init__(self, con):
+        self.con = con
+
+    def belonging(self, segment):
+        """
+        Returns the rate of belonging of each segment per year
+        """
+        query = """
+        SELECT
+            period_end_year AS year,
+            value / 100 AS value,
+            segment
+        FROM belonging
+        WHERE layer = "place"
+        AND segment == "{segment}"
+        GROUP BY year
+        """.format(segment=segment)
+        cur = self.con.cursor()
+        cur.execute(query)
+        rows = rows_to_dicts(cur, cur.fetchall())
+        return rows
--- a/api/metrics/annually.py
+++ b/api/metrics/annually.py
@@ -11,17 +11,17 @@ class AnnuallyMetrics:
 
     def belonging(self, segment):
         """
-        Returns the rate of belonging of each segment per year
+        Returns the rate of belonging of a segment per year
         """
         query = """
         SELECT
-            period_end_year AS year,
+            period_end_year || "-01-01" as week,
             value / 100 AS value,
             segment
         FROM belonging
         WHERE layer = "place"
         AND segment == "{segment}"
-        GROUP BY year
+        GROUP BY week
         """.format(segment=segment)
         cur = self.con.cursor()
         cur.execute(query)
--- a/api/metrics/annually_test.py
+++ b/api/metrics/annually_test.py
@@ -1,0 +1,47 @@
+import sys
+sys.path.append("../")
+
+from api.metrics.annually import AnnuallyMetrics
+from api.utils.testing import create_test_db
+
+
+def test_annual_belonging_rate():
+    belonging_table = [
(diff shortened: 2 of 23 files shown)


fabrego524:
As a side note, while I was testing this particular metric using Cloud9, the data is properly displayed. However, because the data for Rate of Belonging only includes data between the years of 2015 and 2018, having other metrics whose data begins in mid-2018 or later on the timeline means much of the Belonging data will be cut off. As far as I can tell, this means that in order to display Belonging data on the timeline, it has to be the only metric shown. If there was more data for Belonging, it would have been nice to see other metrics displayed alongside it.

app/site/metrics.js
```diff
@@ -140,6 +140,36 @@ export const weeklyMetrics = {
     format: Formatter.numberWithCommas,
     fullFormat: Formatter.numberWithCommas,
   },
+  belonging_rate_all: {
+    name: "Rate of Belonging of the City",
+    units: "of people",
```
vingkan, line 145: This is good, let's keep it like this.

But just something to watch out for, this might confuse visitors:

Screen Shot 2021-07-16 at 2 40 23 PM

The correct interpretation is 52% of Black people in Chicago, but maybe some people might be confused about whether it means 52% of all people in Chicago.

I like the way you have done it, where the population segment is in the metric name and the units is just "of people" so let's keep it that way.

api/metrics/annually_test.py
```diff
@@ -0,0 +1,47 @@
...
+            "layer": "place",
+            "period_end_year": 2017,
+            "segment": "all",
+            "value": 37.1
+        },
+        {
+            "layer": "place",
+            "period_end_year": 2018,
+            "segment": "all",
+            "value": 35.4
```
vingkan, line 32: I would suggest adding a row that does not have segment = "all" to make sure that it actually gets excluded.

app/site/metrics.js
fabrego524: Yeah I realized there was no easy way of maintaining consistency with the other Belonging metrics, so I just went ahead and used the titles used by Chicago Health Atlas for the metric name, and stuck with using the "of people" unit from before. If anything we can change them to match the respective metric name.

api/metrics/annually.py
```diff
@@ -0,0 +1,29 @@
+from api.utils.database import rows_to_dicts
+
+
+class AnnuallyMetrics:
```
vingkan, line 4: Let's rename this and the other locations to be called `YearlyMetrics`.

api/metrics/annually.py
```diff
@@ -0,0 +1,29 @@
...
+        Returns the rate of belonging of a segment per year
+        """
+        query = """
+        SELECT
+            period_end_year || "-01-01" as week,
+            value / 100 AS value,
+            segment
+        FROM belonging
+        WHERE layer = "place"
+        AND segment == "{segment}"
```
vingkan, line 23: Because this is wrapped in quotes, we don't have a risk of SQL injection (e.g., if someone asked for segment = ";drop tables;"). So this should be safe.

vingkan (changes requested):
Excellent work, Fabian! Our first non-weekly metrics for the timeline view.

There are some issues with changing between weekly and yearly metrics, but those are with the Timeline component, not with your code.

I requested two changes before we merge:
- Rename from annually to yearly
- Add another row to the input of your unit test to make sure other segments are not included

api/metrics/annually.py
```diff
@@ -0,0 +1,29 @@
...
+    def __init__(self, con):
+        self.con = con
+
+    def belonging(self, segment):
+        """
+        Returns the rate of belonging of a segment per year
+        """
+        query = """
+        SELECT
+            period_end_year || "-01-01" as week,
```
vingkan, line 18: This is correct, but for some reason, the frontend renders the date as December 31st. I think it's a timezone issue on the frontend, I will look into that and fix.

vingkan (approved):
Excellent work! Let's merge it.

PR #67, opened 2021-07-17
Metric implementation and metric test attempt for Rent Burdened Household

wpj0110:
Might add a few more methods, please check if I'm on the right page.

### DIFF
--- a/api/metrics/rent_burdened.py
+++ b/api/metrics/rent_burdened.py
@@ -1,0 +1,82 @@
+from api.utils.database import rows_to_dicts
+
+
+class RentBurdenedMetrics:
+    """
+    Metrics for rent burdened households data.
+    """
+
+    def __init__(self, con):
+        self.con = con
+
+    def rent_burdened(self, year, segment):
+        """
+        Returns the percent of households who are rent burdened in the area,
+        and the area number.
+        Args:
+            year (int): period ending year to filter by
+            segment (str): population segment to filter by
+        """
+        query = """
+        SELECT
+            area_number,
+            CAST(value AS INTEGER) AS value
+        FROM rent_burdened_households
+        WHERE period_end_year == {year}
+        AND segment == "{segment}"
+        """.format(year=year, segment=segment)
+        cur = self.con.cursor()
+        cur.execute(query)
+        rows = rows_to_dicts(cur, cur.fetchall())
+        return rows
+        
+    def get_max_burdened(self):
+        """
+        Returns the highest percentage of rent burdened households in an area,
+        its area number, and the period.
+        """
+        query = """
+        SELECT
+            max(ROUND(value,2)) AS value,
+            area_number,
+            period_end_year
+        FROM rent_burdened_households
+        """
+        cur = self.con.cursor()
+        cur.execute(query)
+        rows = rows_to_dicts(cur, cur.fetchall())
+        return rows
+    
+    def get_min_burdened(self):
+        """
+        Returns the smallest percentage of rent burdened households in an area,
+        its area number, and the period end year.
+        """
+        query = """
+        SELECT
+            min(ROUND(value,2)) AS value,
+            area_number,
+            period_end_year
+        FROM rent_burdened_households
+        """
+        cur = self.con.cursor()
+        cur.execute(query)
+        rows = rows_to_dicts(cur, cur.fetchall())
+        return rows
+    
+    def average_burden_area(self):
+        """
+        Returns the average percentage of rent burdened households by area
+        across all periods.
+        """
+        query = """
+        SELECT
+            ROUND(avg(value),2) AS "avg value",
+            area_number
+        FROM rent_burdened_households
+        GROUP BY area_number
+        """
+        cur = self.con.cursor()
+        cur.execute(query)
+        rows = rows_to_dicts(cur, cur.fetchall())
+        return rows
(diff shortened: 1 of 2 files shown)


api/metrics/rent_burdened_test.py
```diff
@@ -0,0 +1,77 @@
+import sys
+sys.path.append("../")
+
+from api.metrics.community import CommunityMetrics
```
vingkan, line 4: Looks like your tests are failing because you need to import your new metric class here instead of the community metrics class.

api/metrics/rent_burdened.py
```diff
@@ -0,0 +1,82 @@
...
+        Returns the percent of households who are rent burdened in the area,
+        and the area number.
+        Args:
+            year (int): period ending year to filter by
+            segment (str): population segment to filter by
+        """
+        query = """
+        SELECT
+            area_number,
+            CAST(value AS INTEGER) AS value
```
vingkan, line 23: If your metric is for a percentage or rate, return the value as a value between 0 and 1. We will take care of formatting it as a percentage on the frontend.

If all of the metrics that are rates or percentages return values between 0 and 1 it makes it easier to display color scales and to compare different metrics on the same axis when making data visualizations.

api/metrics/rent_burdened.py
```diff
@@ -0,0 +1,82 @@
...
+        FROM rent_burdened_households
+        WHERE period_end_year == {year}
+        AND segment == "{segment}"
+        """.format(year=year, segment=segment)
+        cur = self.con.cursor()
+        cur.execute(query)
+        rows = rows_to_dicts(cur, cur.fetchall())
+        return rows
+        
+    def get_max_burdened(self):
```
vingkan, line 33: Great work, William! You went over and above adding more metrics.

api/metrics/rent_burdened.py
```diff
@@ -0,0 +1,82 @@
...
+            area_number,
+            period_end_year
+        FROM rent_burdened_households
+        """
+        cur = self.con.cursor()
+        cur.execute(query)
+        rows = rows_to_dicts(cur, cur.fetchall())
+        return rows
+    
+    def average_burden_area(self):
```
vingkan, line 67: This is an interesting idea for a metric. Averaging percentages might be misleading because each percentage represents a different year with a different denominator, so we don't have a way to weight the averages. For example, suppose that between two years, the number of rent burdened households stays the same, but in the second year, some new houses are finished and non rent burdened people move in. The denominator of total households increases, causing the percentage of rent burdened to go down, and causing the average to go down.

You don't have to change it, but just wanted to point this out.

api/metrics/rent_burdened_test.py
```diff
@@ -0,0 +1,77 @@
...
+            "./pipeline/load/rent_burdened_households.sql"
+        ],
+        tables={
+            "rent_burdened_households": rent_burdened_households_table
+        }
+    )
+
+    metric = RentBurdenedMetrics(con)
+    
+    assert metric.rent_burdened(year=2019, segment="all") == [
```
vingkan, line 59: Good unit tests! There are pros and cons to testing many functions in the same test case with the same input data:

- Pro: you only have to come up with one dataset
- Pro: if the dataset is simple, the reader can remember it when looking at each test
- Con: if one of the methods changes, you might have to rework the data while making sure you don't break the other tests
Con: if the dataset is complicated, the reader may have a harder time telling which inputs affect the output of each function

I think the two pros apply and the two cons are not as much of a worry in this case, so you can keep this test the way it is, just wanted to comment on this.

vingkan (changes requested):
Looking good, William! When you get the unit tests to pass you can merge.

api/metrics/rent_burdened.py
wpj0110: Should I just add an sql script to divide my values by 100 before I load them into the database?
Referring to` pipeline/load/rent_burdened_households.sql`

api/metrics/rent_burdened_test.py
wpj0110: Thanks for finding this!

api/metrics/rent_burdened.py
vingkan: It's okay to do the division here in the metric! That's what the other metrics do.

But if you want to change what's in the database, I would do the division in your transform Python script. The load SQL is just for defining the table schema, not modifying the data.

api/metrics/rent_burdened_test.py
```diff
@@ -0,0 +1,76 @@
...
+    assert metric.get_max_burdened() == [
+        { 'value': 0.562945422971736, 'area_number': 2, 'period_end_year': 2019}
+    ], "Should have max result of 0.5777931000000001, area_number of 2, and period end year of 2019."
+    
+    assert metric.get_min_burdened() == [
+        { "area_number": 3, "value": 0.339090007521057, "period_end_year": 2017}
+    ], "Should have min result of 32 and area_number of 3."
+
+    assert metric.average_burden_area() == [
+        {'avg value': 0.47178547580318553, 'area_number': 1},
```
vingkan, line 73: Float division can be unstable in Python, so I would suggest doing something like this:

```python
# input data
        {
            ...
            "value": 33.9
        }
...
# test assertion
    assert ... == [
        {'avg value': 33.9 / 100, ...},
      ]
```

vingkan (approved):
Looks good! I left one more nitpick about the decimal places in the unit tests, up to you if you want to change that before you merge.

wpj0110:
Oh no I merged before reading your comment Vinesh! My bad, I'll try to make the changes.

PR #69, opened 2021-07-17
Disabilities metrics tests

MariaGarciaCS:
The tests for the disability rate metric all pass, but two other tests do not. 

```
pipeline/tests/database_test.py::test_table_counts FAILED
pipeline/tests/database_test.py::test_belonging FAILED 
```

I updated my repo following the thread in "#help" between Vinesh and Asude, but the tests continue to fail. The only file I modified is community_test.py

### DIFF
--- a/api/metrics/community_test.py
+++ b/api/metrics/community_test.py
@@ -232,6 +232,13 @@ def test_disabilities():
             "segment": "all",
             "value": 8.46
         },
+        {
+            "area_number": 4,
+            "period_end_year": 2015,
+            "segment": "other",
+            "value": 13.37
+        },
+        
     ]
     con, cur = create_test_db(
         scripts=[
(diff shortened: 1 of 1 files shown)


api/metrics/community_test.py
```diff
@@ -210,4 +210,47 @@ def test_belonging():
...
+            "period_end_year": 2016,
+            "segment": "all",
+            "value": 11.02
+        },
+        {
+            "area_number": 3,
+            "period_end_year": 2016,
+            "segment": "all",
+            "value": 8.46
+        },
```
vingkan, line 234: Good test! I would add one more row with a segment that is not "all" to show that the segment filter works.

vingkan (changes requested):
Thanks for adding this test! GitHub says your tests pass and I pulled your branch to run tests on my Cloud9 and they pass there as well.

I suggested one change to the input data for your test. After you make that change you can merge.

MariaGarciaCS:
Test for segment filter has been added. All tests pass.

vingkan (approved):
Good to merge! Thanks for updating the test.

Based on the diff, it seems like your PR changed the compressed database file even though it shouldn't have. The tests pass, so it's probably an safe change, and we don't have an easy way to fix that after it's been committed. So let's go ahead and merge it for now.

PR #70, opened 2021-07-17
New options for run app: hotswap API and serve static site.

vingkan:
New options for running the app, added to the common commands doc.

- **If you are doing frontend work only,** you can use `run app 9 api prod` to run your app in development mode, while using the production API. This means you do not have to run the API on your Cloud9.
- **If you are doing backend work only,** you can use `run app 9 static` to serve a static version of the app, while using your Cloud9 API. This means the frontend will not rebuild on changes, but will be faster to navigate between pages.

Here is a screencast showing a demo: https://www.loom.com/share/7751ee3dd41142e2a4c25159b88d1dce

### DIFF
--- a/api/dev.sh
+++ b/api/dev.sh
@@ -7,7 +7,7 @@ source .venv/bin/activate
 pip3 install -r requirements.txt
 
 # Refresh IP address for Cloud9 instance
-if [ -n "$1" ]; then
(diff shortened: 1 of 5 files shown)


divyashie (approved):
LGTM! 👍🏽

PR #72, opened 2021-07-18
Traffic intensity transform

sperveen4:
Transformed traffic_intensity dataset in py file

### DIFF
--- a/pipeline/transform/traffic_intensity.py
+++ b/pipeline/transform/traffic_intensity.py
@@ -1,0 +1,30 @@
+import argparse
+import pandas as pd
+from timeit import default_timer as timer
+
+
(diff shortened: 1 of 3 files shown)


pipeline/transform/traffic_intensity.py
```diff
@@ -14,8 +14,7 @@
 df = pd.read_csv(args.input_file)
 
 # Split the period (e.g. 2015-2019) into start and end year
```
Andrew-Tolentino, line 16: Just would remove this comment since for this dataset we found out that there are no period ranges like the population dataset

Andrew-Tolentino (approved):
Awesome job! Really small suggestion of just removing a comment

Will approve so you can merge.

PR #74, opened 2021-07-19
CTA train ridership transform step

ashklover:


rbrooks6 (approved):
Nice job!

pipeline/transform/cta_train_ridership.py
```diff
@@ -0,0 +1,23 @@
...
+cli = argparse.ArgumentParser(description="Transform CTA L Train ridership data.")
+cli.add_argument("--input_file", help="File path to read CTA L Train ridership data from.")
+cli.add_argument("--output_file", help="File path to write results to.")
+args = cli.parse_args()
+
+start = timer()
+
+# Read file
+df = pd.read_csv(args.input_file)
+
```
vingkan, line 15: @ashklover @rbrooks6 In a future PR, I think it would be a good idea to add a column for the week of the year. That way we can aggregate ridership by week, not just by day, and compare it to other kinds of transit that do not have daily counts.

There is a helper function called `extract_data_portal_dates()` to do this:
- [Example in rideshare code](https://github.com/scarletstudio/transithealth/blob/main/pipeline/transform/weekly_rideshare.py#L28)
- [Documentation in common data transformations](https://github.com/scarletstudio/transithealth/blob/main/docs/pages/transformations.md#parse-date)

vingkan: Also, I realize that helper function assumes the date will be in %Y-%m-%d format with a timestamp, but for your dataset the dates are in %m-%d-%Y with no timestamp.

You could:
- Create a new method that supports the date format in this dataset
- Refactor the existing method to support multiple date formats

PR #76, opened 2021-07-19
Transform Sidewalk Cafe Data

asudeozturk:
Modified Makefile and added transform/sidewalk_cafe.py


morleegan (approved):

pipeline/transform/sidewalk_cafe.py
```diff
@@ -0,0 +1,76 @@
...
+#Drop lat lon columns (too many null)
+df.drop([
+    "latitude",
+    "longitude",
+], axis=1, inplace=True)
+
+
+#Create new column "exists". Assign 0 for expired permits, 1 for existing permits
+today = date.today().strftime("%Y-%m-%d")
+df['exists'] = np.where( (df['issued_date'] < today ) & (today < df['expiration_date']), 1, 0)
```
vingkan, line 39: I would recommend that we check whether the permit is existing or expired in the API rather than in the offline pipeline, that way we can use the current date instead of the date when we ran our pipeline.

Or was there another reason for adding the exists column in the offline pipeline?

pipeline/transform/sidewalk_cafe.py
```diff
@@ -0,0 +1,76 @@
...
+df = convert_zip_to_community_area_proportionally(
+    df_zip_to_area,
+    df,
+    cols_to_convert=["exists"],
+    zip_col="zip",
+    verbose=True
+)
+
+
+#Add community area names based on area numbers
```
vingkan, line 57: This join can be done in the API. In fact, the community (scatter) view on the website automatically does this join.

vingkan (commented):
Great work @asudeozturk! You added a lot of transformations. I left some comments on some transformations that I think can be done in the API instead of the offline pipeline.

I didn't realize that there were so many nulls for longitude and latitude. Good catch! You can definitely create a timeline metric for sidewalk cafes over time. But we'll have to think more about how to do a community area metric. Here are some notes I have:

- A longitude/latitude point can belong to exactly one community area
- However, a zip code can overlap with one or more community areas
- The `convert_zip_to_community_area_proportionally()` method will take the columns you give it and multiply its values by the fraction of the population from the zip code that lives in each community area it overlaps with
- Population fraction may or may not be a good estimate of which community area the sidewalk cafe is located in

So we have a bunch of options:

A. Don't do any community area-based metrics, just time-based metrics
B. Use the street address to "geocode" and get longitude/latitude coordinates for each sidewalk cafe and use the coordinates to get community area
C. In the API, write a metric to get number of cafes per zip code, join that to a table of conversion factors, and then get an estimate of number of cafes per community area based on population
D. Add a new view to our site to show metrics by zip code, where some metrics will only be available by zip or by community area but not both

For now, I recommend option A, so that you can keep moving and add a metric to the API. Any of the other options would be worthy of a stretch project, so if you are interested, we could revisit that in week three.

CC: @morleegan

PR #77, opened 2021-07-19
Frontend disabilities

MariaGarciaCS:
Added question looking at the percentage of residents with a disability in a selected area. 

Changes made since pair programming with Vinesh
-Changed descriptions
-Location is now Burnside
-Chart Colors


app/components/questions/ResidentsWithDisabilities.js
```diff
@@ -0,0 +1,145 @@
...
+  Tooltip,
+  Legend,
+} from 'recharts'
+import { FailureNotification } from '../../components/Notification'
+import { Color } from '../../site/theme'
+import { Formatter } from '../../site/metrics'
+
+const ENDPOINT = `${process.env.NEXT_PUBLIC_API}/community/metrics`
+
+const TRANSIT_MODES = {
```
vingkan, line 17: We no longer need this variable, so you can delete lines 17-34.

api/endpoints/timeline.py
```diff
@@ -26,7 +26,8 @@ def make_blueprint(con):
         "yearly_belonging_rate_W": lambda: yearly_metric.belonging("W"),
         "yearly_belonging_rate_B": lambda: yearly_metric.belonging("B"),
         "yearly_belonging_rate_A": lambda: yearly_metric.belonging("A"),
-        "yearly_belonging_rate_H": lambda: yearly_metric.belonging("H")
+        "yearly_belonging_rate_H": lambda: yearly_metric.belonging("H"),
```
vingkan, line 29: Oops, we forgot to commit your changes that you were working on before moving to a new branch. Can you make a branch that has just the changes to the frontend?

vingkan (changes requested):
I left two comments before we can merge this.

app/components/questions/ResidentsWithDisabilities.js
MariaGarciaCS: These lines have been removed

api/endpoints/timeline.py
MariaGarciaCS: These changes have been removed and exist on the timeline_disabilities branch

vingkan: Excellent! it looks like you moved the metric implementation, but not the test.

Since the test is still in this branch, it is failing.

Once you move the test, we can merge!

vingkan:
We will also need to merge main into your branch because Fabian also added a question.

api/endpoints/timeline.py
MariaGarciaCS: Test was removed, all tests pass now.

vingkan (approved):
LGTM!

PR #79, opened 2021-07-19
Frontend Belonging

fabrego524:
Implemented a Frontend example for the Belonging Dataset, includes a basic pie chart. Also includes minor changes among yearly.py and yearly_test.py to maintain consistency with naming.

PR #83, opened 2021-07-20
Improve Frontend Belonging Formatting

fabrego524:
This was first attempted in I believe both PR #79 and #77 but couldn't be figured out, so I thought it'd be good to find out a solution to the unformatted values being used by the page.

I managed to use Recharts and React API and examples to create custom styling for both piechart label and tooltip, including the change from the unformatted decimal to formatted percentage. I tried doing this while keeping the footprint of the actual piechart code to a minimum, meaning I had to use various `const`s to define the behavior I wanted, which could then be easily plugged into the syntax for the piechart.

The code is a bit inconsistent with itself, the manner in which Recharts interacts with the provided data is a bit confusing. For the custom tooltip here, I couldn't figure out a way to use the previously added `valueFormatted`, so I instead performed the same operation that our Formatter does with `percentWithOneDecimal`. Perhaps there is a better way of handling this that I'm not aware of.

https://github.com/scarletstudio/transithealth/blob/8d8d221e7763927fe8b59f3c1f5059e7c3d2ed04/app/components/questions/BelongingRates.js#L73-L79

On the other hand, the code for the custom label is a bit easier to work with. It's able to calculate the position of the labels themselves using a few math operations and otherwise consists of using the existing `color` and `valueFormatted` that was set in the data to return a simple piece of text, which should make it useful if we ever manage to add a menu of some sort to change the area being displayed on the page.

Unfortunately, I'm not sure how well this would work with other data visualization types. However, what I've read from documentation and sources online indicates it should be similarly possible.

### DIFF
--- a/app/components/questions/BelongingRates.js
+++ b/app/components/questions/BelongingRates.js
@@ -21,6 +21,16 @@ const EMPTY_AREA = {
   belongingRateFormatted: "?",
 };
 
+const RADIAN = Math.PI / 180;
+const renderCustomizedLabel = ({ percent, index }) => {
+
+  return (
+    <text>
+      {`${(percent * 100).toFixed(1)}%`}
+    </text>
+  );
+};
+
 function transformData(response, error, selectedArea) {
   if (!response || error) {
     return {
@@ -40,14 +50,17 @@ function transformData(response, error, selectedArea) {
     {
       label: "Feels Belonging",
       value: areaData.belonging_rate_2018,
+      valueFormatted: Formatter.percentWithOneDecimal(areaData.belonging_rate_2018),
       color: Color.Indigo
     },
     {
       label: "Does Not Feel Belonging",
       value: 1 - areaData.belonging_rate_2018,
+      valueFormatted: Formatter.percentWithOneDecimal(1 - areaData.belonging_rate_2018),
       color: Color.Salmon
     }
     ]
+
   
   return {
     chartData: data,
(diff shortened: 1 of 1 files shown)


app/components/questions/BelongingRates.js
```diff
@@ -57,6 +60,48 @@ function transformData(response, error, selectedArea) {
...
+    borderWidth: "thin"
+  };
+  
+  
+  const CustomTooltip = ({ active, payload, label }) => {
+  if (active && payload && payload.length) {
+    // Changes the tooltip to display the formatted percentage rather than a decimal
+    return (
+      <div className="custom-tooltip" style={tooltipStyle}>
+        <p className="label">{`${payload[0].name} : ${(payload[0].value * 100).toFixed(1)}%`}</p>
```
vingkan, line 78: Instead of:

```js
${(payload[0].value * 100).toFixed(1)}%
```

Use:

```js
${payload[0].valueFormatted}
```

app/components/questions/BelongingRates.js
```diff
@@ -57,6 +60,48 @@ function transformData(response, error, selectedArea) {
 
 function ExamplePieChart(props) {
   const { data } = props
+  
+  
+  const tooltipStyle = {
```
vingkan, line 65: Good choice to move this into a separate constant!

app/components/questions/BelongingRates.js
```diff
@@ -57,6 +60,48 @@ function transformData(response, error, selectedArea) {
...
+        <p className="label">{`${payload[0].name} : ${(payload[0].value * 100).toFixed(1)}%`}</p>
+      </div>
+    );
+  }
+  return null;
+  };
+  
+  
+  const RADIAN = Math.PI / 180;
+  const renderCustomizedLabel = ({ cx, cy, midAngle, innerRadius, outerRadius, color, valueFormatted, index }) => {
```
vingkan, line 87: Believe it or not, as far as I know, this is the intended way in Recharts to customize labels and tooltips. Great job researching and learning it!

Each page can have its own custom labels and custom field components. If we have resuable custom components, we can move them into common files so others can import them.

vingkan (approved):
Great work Fabian! Very glad to see you were able to figure out how to do this on your own.

Requested a few small changes, then this is good to go!

vingkan:
Well done, Fabian! I pulled your branch and ran it on my Cloud9, looks great.

app/components/questions/BelongingRates.js
fabrego524: Confusingly, this change doesn't work as expected. I might have to look into the way the payload works, but on the page it displays as `undefined` instead of what you'd expect from `valueFormatted`

image

vingkan: Oh, haha I logged `payload` and it looks like there's another level to go into:

Screen Shot 2021-07-20 at 3 06 33 PM

So we can do:

```js
const entry = payload[0].payload;
...
${payload.valueFormatted}
```

app/components/questions/BelongingRates.js
```diff
@@ -81,7 +126,7 @@ function ExamplePieChart(props) {
             />
           ))}
         </Pie>
-        <Tooltip />
+        <Tooltip content={<CustomTooltip />} />
         <Legend
           layout="horizontal"
           verticalAlign="top"
```
vingkan, line 132: Delete the console.log on line 146.

app/components/questions/BelongingRates.js
fabrego524: I think I've figured it out. Instead of
```javascript
${payload[0].valueFormatted}
```
we need to use 
```javascript
${payload[0].payload.valueFormatted}
```
This is a bit weird but it works from what I can see.

PR #87, opened 2021-07-21
Remove venv reference from new_endpoint.md

rbrooks6:
remove step that manually activates virtual environment


vingkan (approved):
Thanks for catching this!

PR #89, opened 2021-07-21
Added transform operations

camilo-cortes:


rbrooks6:
This PR is still in the draft step. We are working on adding the transform file to complete this

pipeline/Makefile
```diff
@@ -301,11 +314,11 @@ data/extracted/cta_train_ridership.csv:
 		--output_file="data/extracted/cta_train_ridership.csv"
 		
 
-data/extracted/daily_bus_rideship.csv:
+data/extracted/daily_bus_ridership.csv:
     python3 extract/from_data_portal.py \
-        --json_url="$(PORTAL_BUS_RIDESHIP)" \
-        --soql_file="extract/daily_bus_rideship.sql" \
-        --output_file="data/extracted/daily_bus_rideship.csv"
+        --json_url="$(PORTAL_BUS_RIDERSHIP)" \
```
rbrooks6, line 319: The spacing in this one looks different in comparison to the other steps under this section of the make file. Try to backspace all of the indented lines after line 317 to be in line with that line and then use tab to indent them

pipeline/transform/daily_bus_ridership.py
```diff
@@ -0,0 +1,26 @@
...
+cli.add_argument("--output_file", help="File path to write results to.")
+args = cli.parse_args()
+
+start = timer()
+
+# Read file
+df = pd.read_csv(args.input_file)
+
+# Parse dates and add week column
+#df = extract_data_portal_dates(df, col="date", prefix="")
```
rbrooks6, line 18: uncomment this

pipeline/load/daily_bus_ridership.sql
```diff
@@ -0,0 +1,7 @@
+DROP TABLE IF EXISTS daily_bus_ridership;
+CREATE TABLE daily_bus_ridership(
+    route INTEGER,
+    date DATE,
+    daytype TEXT,
+    rides INTEGER
```
rbrooks6, line 6: let's add a week column to our table to incorporate the week that we added on the transform step
it would look like this:
```
rides INTEGER,
week TEXT
```

pipeline/transform/Untitled.ipynb
```diff
@@ -0,0 +1,192 @@
+{
```
rbrooks6, line 1: we don't want to commit this, but you can keep it locally by doing the following: git rm --cached pipeline/transform/Untitled.ipynb. 

This will stop git from tracking it, without removing it from your local files

run
```diff
@@ -0,0 +1,92 @@
+#!/bin/bash
```
rbrooks6, line 1: Hmm is this a duplicate of the run.sh script ? We should be able to remove this with git rm run

pipeline/transform/Untitled.ipynb
rbrooks6: From the transithealth directory:
git mv [from] [to]
git mv pipeline/transform/Untitled.ipynb notebooks/ccortes2/TestNotebook.ipynb

rbrooks6 (approved):
Looks great!

rbrooks6:
Looks like that check failed. Try doing that run pipeline-quick again and then push once it finishes

rbrooks6 (approved):
:+1:

PR #91, opened 2021-07-21
One last merge conflict. Backend to Frontend for Rent Burdened Households data

wpj0110:
Hopefully there's no more merge conflict.... and I finally have my stuff show up to the front end!


app/site/metrics.js
```diff
@@ -107,6 +107,54 @@ export const communityMetrics = {
     format: Formatter.percentWithNoDecimal,
     fullFormat: Formatter.percentWithOneDecimal,
   },
+  rent_burdened_2015: {
+    name: "2015 Rate of Rent Burdened Households",
+    units: "of households",
+    format: Formatter.percentWithOneDecimal(),
```
vingkan, line 113: Delete the parentheses from all of the format and fullFormat lines.

The difference is:

- If you include the parentheses, that calls the method right now
- If you leave out the parentheses, that gives a variable for the method

We want the latter, that way we can pass the format function to whatever code will actually call it and apply the formatting.

api/endpoints/community.py
```diff
@@ -27,6 +29,14 @@ def make_blueprint(con):
         "disability_rate_2019":lambda: metric.disability_rate(year=2019, segment="all"),
         "belonging_rate_2017": lambda: metric.belonging(year=2017, segment="all"),
         "belonging_rate_2018": lambda: metric.belonging(year=2018, segment="all"),
+        "rent_burdened_2015": lambda: metric_rbu.rent_burdened(year=2015, segment="all"),
+        "rent_burdened_2016": lambda: metric_rbu.rent_burdened(year=2016, segment="all"),
+        "rent_burdened_2017": lambda: metric_rbu.rent_burdened(year=2017, segment="all"),
```
vingkan, line 34: For now, let's just load the years 2017-2019, that should be plenty for this view.

Later if you want, you can add a timeline metric that shows the value over the years.

api/endpoints/community.py
```diff
@@ -27,6 +29,14 @@ def make_blueprint(con):
...
+        "rent_burdened_2018": lambda: metric_rbu.rent_burdened(year=2018, segment="all"),
+        "rent_burdened_2019": lambda: metric_rbu.rent_burdened(year=2019, segment="all"),
+        "rent_max": lambda: metric_rbu.rent_max_burdened(),
```
vingkan, line 37: The min and max metrics are missing a `GROUP BY area_number` in their query. Also you can remove the `period_end_year` from those two metrics.

vingkan (changes requested):
I pulled your branch to try it out and found some fixes. Let me know if you need help with the merge conflicts.

app/site/metrics.js
wpj0110: Deleted!

vingkan (approved):
LGTM after we fix the unit tests for the min and max metrics.

PR #94, opened 2021-07-21
Frontend Selector for Belonging Question

fabrego524:
Implemented a simple selector with a list of all community areas in Chicago. Uses a new function `AreaSelector` that generates every option, and now has a React `useState` to change the page upon a new selection.

An interesting case that may have to be handled in the future is the Burnside Community Area. As was encountered during the metric implementations for belonging, there is no data for Burnside. This is reflected on the page when it is selected.

image


vingkan (approved):
Excellent! Good to merge.

If you figure out how to handle the Burnside case, that can be addressed in a separate PR.

PR #96, opened 2021-07-22
edited transform file to create week of the year columns

ashklover:


vingkan (commented):
@ashklover this change looks good. How about you also load the new week column in this PR as well?

Add the new column to your schema and re-run the pipeline to get new compressed DB files.

You'll need to merge `main` into your branch and re-run the pipeline anyway, since GitHub says that `pipeline/compressed/db-part-aa` has a conflict between `main` and your branch.

PR #103, opened 2021-07-22
Fix bug where changing metric causes scatter view tooltip to display NaN or crash the app

### LINKED ISSUE #92: Scatter view custom tooltip tries to format data before it is available (flaky)
## Description

Sometimes, the community scatter view tries to apply the format function to data for a metric before that metric is actually received from the API response, which causes an error.

@wpj0110 and I noticed this while playing around with the app.

## To Reproduce

Warning: very flaky, hard to reproduce because it depends on timing.

- Hover over a point on the scatter plot to pull up the custom tooltip
- Start rapidly changing metrics from the selectors until you get this bug lol

## Expected Behavior

The scatter view should not try to format the metric until the data is loaded.

Maybe we need to have an intermediate loading state for the component.

vingkan:
Fixes bug #92 

This screencast explains the cause of the bug (stale data in the custom tooltip) and the solution (detect when the data doesn't have the new metrics and tell the tooltip not to display).

https://www.loom.com/share/88c9b582ac8447cb94ed529cc3c206cd

Thanks @fabrego524 and @wpj0110 for helping find this bug!

Clarifying question from the video:

> In the video, you said that the data inside the custom tooltip is stale. Then why do the metrics in the tooltip change but not the payload?

> Yes, that is something I should have explained in the video. The metrics change because the metrics are passed into the custom tooltip as a property from the parent component. However, the payload is managed by the Recharts library that we use for data visualizations and is not under the control of either component. Recharts does not update the payload while the tooltip is open, so we get stale data. I couldn't find out online whether this is the intended behavior of Recharts or if this is bug.

### DIFF
--- a/app/components/CommunityScatterExplorer.js
+++ b/app/components/CommunityScatterExplorer.js
@@ -59,11 +59,17 @@ function getGeoMapMetric(metrics, selectedMetric, minAlpha=0.05) {
   return metricMapData;
 }
 
-function CustomToolTip({ active, payload, label, metrics, selectedPayload }) {
-  if (!active || !payload || payload.length === 0) {
+function CustomToolTip({ active, payload, label, metrics }) {
+  // The ?. is called "optional chaining"
+  // If the property after the ?. is not found, we will return undefined
+  // and stop trying to evaluate the remaining nested properties.
+  const d = payload?.[0]?.payload || {};
+  // Check if any of the metrics are not in the data, which happens if the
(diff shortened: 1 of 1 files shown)


fabrego524 (approved):
It seems like the solution works solidly, all of my attempts to recreate the issue don't result in the bug anymore, since the tooltip stops displaying now as the scatter plot changes. LGTM.

Issue #61, opened 2021-07-16
Timeline view shows dates incorrectly in tooltip.


vingkan:
- For example, for yearly metrics, the date 2019-01-01 gets displayed as December 30, 2018.
- The x-axis seems to display the dates correctly, albeit smush together.
- I think this is a timezone issue on the frontend, the backend seems correct since it is returning a formatted string.

vingkan:
Steps to reproduce:
- Go to the timeline view
- Remove the default metrics
- Click one of the rate of belonging metrics
- Hover over one of the data points
- The date should show as January 1, 2016 (or whatever the year number is
- However, if your timezone is behind GMT, then you will see the incorrect date of December 31, 2015

PR #104, opened 2021-07-22
Hack to fix dates on timeline tooltip.

vingkan:
Fixes bug #61 

# Try the Fix

To try this change out in your Cloud9:

```bash
cd transithealth
git fetch
git checkout fix-timeline-dates
run app 9 api prod
```

- Go to the timeline view
- Remove the default metrics
- Add one of the belonging metrics
- Hover over one of the data points and check if the date is correct

# Before (Bug)

Screen Shot 2021-07-22 at 12 26 19 AM

# After (Fixed)

Screen Shot 2021-07-22 at 12 25 49 AM

### DIFF
--- a/app/components/TimelineExplorer.js
+++ b/app/components/TimelineExplorer.js
@@ -51,14 +51,18 @@ async function getTimelineMetrics(metrics) {
 }
 
 function CustomToolTip({ active, payload, label, metrics, selectedPayload }) {
-  if (!active || !payload || payload.length === 0) {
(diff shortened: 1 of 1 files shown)


vingkan:
Requesting @rbrooks6's review on this PR so someone who is not in the central timezone can try out the fix!

vingkan:
I changed my computer timezone to Alaska and mainland China time to confirm that this change works!

PR #108, opened 2021-07-22
Added 2 more years for Median Income metric

fabrego524:
Went ahead and implemented 2 additional years for the Median Income metric, 2017 and 2018. This will expand the data a bit for the scatter plot and provide a bit more data for some questions in the future.


vingkan (approved):
LGTM!

PR #111, opened 2021-07-23
updated transformed taxi data

alkatoutl:
I decided to just fill null values in pickup and dropoff community area columns with 0's rather than delete the entire row


vingkan (approved):
Transform change looks good. The load SQL file is empty, did you mean for that?

alkatoutl:
> Transform change looks good. The load SQL file is empty, did you mean for that?

Yeah, I'm working on the load SQL file right now

PR #112, opened 2021-07-23
loaded table

alkatoutl:


vingkan (changes requested):
Great work Leilah!

Can you add a test to `pipeline/tests/database_test.py` that checks the number of rows in the table? That way the test will fail if someone hasn't used `run pipeline-quick` to get your new dataset. There should be one function where all the table size tests are.

Also, how long does it take to run the extract and transform steps? Just wanted to know if we should think about one or both of these in the future:

1) loading an aggregated version (like trips per day) rather than all trips
2) storing the extracted and transformed taxi datasets in the archive so others don't have to rerun them

alkatoutl:
> Great work Leilah!
> 
> Can you add a test to `pipeline/tests/database_test.py` that checks the number of rows in the table? That way the test will fail if someone hasn't used `run pipeline-quick` to get your new dataset. There should be one function where all the table size tests are.
> 
> Also, how long does it take to run the extract and transform steps? Just wanted to know if we should think about one or both of these in the future:
> 
> 1. loading an aggregated version (like trips per day) rather than all trips
> 2. storing the extracted and transformed taxi datasets in the archive so others don't have to rerun them

sure, I just added the test and will push it after this reply. The extracting takes 14 seconds and the transforming takes 6 seconds, is that too long? I was confused on how I could condense all trips into trips per day since they all have different trip_minutes, payment_types, etc.

vingkan (approved):
Test added, LGTM!

Right now, we are only loading 100,000 taxi ride records from 2021. This means that metrics we make from the current data will be misleading. But I would rather that you get to start working on the API and frontend than load the rest of the data now

This is a large dataset, so we will have to try some different approaches to load it in an efficient way, like we did for the rideshare data. @divyashie Later, I can pair program with @alkatoutl on how to do this. We may also be able to get the 2019-2021 data as well.

PR #114, opened 2021-07-23
added taxitrips table size test

alkatoutl:


vingkan:
Can you move this to the same branch as your other PR? Since it's a one line change, I would recommend copying it, switching to your other branch, pasting it, the committing to that branch.

PR #116, opened 2021-07-23
Frontend Belonging Solution for Burnside

fabrego524:
Implemented a simple check to determine if a particular community area returns `NaN` for its value which is passed to both the pie chart and a new `function generateMessage` which will change the text on the page in case the community area has no present data.

One thing of note is that this particular section of code feels a bit clunky for what it does. This specifically looks at the belonging rate of the area whose data is being transformed, and sees if it is `NaN`. There may be some way to check the value directly, however, I was only able to figure out how to check the formatted string instead.

https://github.com/scarletstudio/transithealth/blob/2a039b783069850ce59a8cf615e1025dfeb1747e/app/components/questions/BelongingRates.js#L53-L60

The new function ensures that the parent component isn't responsible for generating this message. It allows for other types of checks such as data from different periods or from different populations. I'm not confident that it's the best practice, but I'm sure it will be useful.

https://github.com/scarletstudio/transithealth/blob/2a039b783069850ce59a8cf615e1025dfeb1747e/app/components/questions/BelongingRates.js#L171-L189

A similar check is added to the piechart from Recharts to hide the element when there is no data for an area. This was the simplest method used in the React documentation, but it's not exactly the cleanest in terms of appearance.

https://github.com/scarletstudio/transithealth/blob/2a039b783069850ce59a8cf615e1025dfeb1747e/app/components/questions/BelongingRates.js#L70-L74

Lastly, I made a simple change to the name of this sample Question to be more clear about its contents.

### DIFF
--- a/app/components/questions/BelongingRates.js
+++ b/app/components/questions/BelongingRates.js
@@ -59,8 +59,9 @@ function transformData(response, error, selectedArea) {
 
 function ExamplePieChart(props) {
   const { data } = props
+  const { areaData } = props
   
-  if(!data[0].value){
+  if(!data?.[0]?.value){
     return null;
   }
   
--- a/app/components/questions/BelongingRates.js
+++ b/app/components/questions/BelongingRates.js
@@ -59,7 +59,6 @@ function transformData(response, error, selectedArea) {
 
 function ExamplePieChart(props) {
   const { data } = props
-  const { areaData } = props
   
   if(!data?.[0]?.value){
     return null;
--- a/app/components/questions/BelongingRates.js
+++ b/app/components/questions/BelongingRates.js
@@ -49,14 +49,14 @@ function transformData(response, error, selectedArea) {
       valueFormatted: Formatter.percentWithOneDecimal(1 - areaData.belonging_rate_2018),
       color: Color.Salmon
     }
-    ]
-    
+  ]
+
   const dataFlag = (() => {
   if (areaData.belongingRateFormatted==="NaN%")
     return 1
   else
     return 0
-})();
+  })();
   
   return {
     chartData: data,
@@ -175,7 +175,7 @@ function GenerateMessage(props) {
       <p>Waiting for data...</p>  
     )
   }
-  
+
   if(props.dataFlag){
     return (
       <span>There is no data for {props.areaData.name}</span>
--- a/app/components/questions/BelongingRates.js
+++ b/app/components/questions/BelongingRates.js
@@ -50,17 +50,27 @@ function transformData(response, error, selectedArea) {
       color: Color.Salmon
     }
     ]
-
+    
+  const dataFlag = (() => {
+  if (areaData.belongingRateFormatted==="NaN%")
+    return 1
+  else
+    return 0
+})();
   
   return {
     chartData: data,
     areaData,
+    dataFlag,
   }
 }
 
 function ExamplePieChart(props) {
   const { data } = props
   
+  if(props.dataFlag){
+    return null;
+  }
   
   const tooltipStyle = {
     backgroundColor: "white", 
@@ -156,7 +166,25 @@ function AreaSelector(props) {
         ))}
       </select>
     </div>
-  )
+  );
+}
+
+function GenerateMessage(props) {
+  if(!props.areaData){
+    return (
+      <p>Waiting for data...</p>  
+    )
+  }
+  
+  if(props.dataFlag){
+    return (
+      <span>There is no data for {props.areaData.name}</span>
+    )
+  } else {
+    return (
+      <span>Based on the data from 2018, {props.areaData.belongingRateFormatted} of people in {props.areaData.name} agree or strongly agree that they feel a sense of belonging in their community</span>
+    )
+  }
 }
 
 export default function BelongingRates(props) {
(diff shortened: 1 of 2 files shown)


app/components/questions/BelongingRates.js
```diff
@@ -156,7 +166,25 @@ function AreaSelector(props) {
...
+}
+
+function GenerateMessage(props) {
+  if(!props.areaData){
+    return (
+      <p>Waiting for data...</p>  
+    )
+  }
+
+  if(props.dataFlag){
```
vingkan, line 179: Instead of putting this in data flag, we could put the check here.

Also, instead of checking:

```js
areaData.belongingRateFormatted==="NaN%"
```

We can check the non-formatted field:

```
!areaData.belonging_rate_2018
```

Note the not `!` which will check if the value is null or undefined. This will also exclude `0`, which might be a legitimate value, so if you want to cover that case you can do:

```js
const missingData = !areaData.disability_rate_2018 && areaData.disability_rate_2018 !== 0
if (missingData) ...
```

You can also "destructure" the props that way you don't have to do `prop.*` everytime:

```js
const { areaData } = props; 
```

Or:

```js
function GenerateMessage({ areaData }) {
```

app/components/questions/BelongingRates.js
```diff
@@ -49,18 +49,28 @@ function transformData(response, error, selectedArea) {
...
+  if (areaData.belongingRateFormatted==="NaN%")
+    return 1
+  else
+    return 0
+  })();
   
   return {
     chartData: data,
     areaData,
+    dataFlag,
```
vingkan, line 64: See comment below, for something like this, we can check it when the component renders, rather than adding it to the data transform.

Also, the syntax here of making an anonymous function and calling it is unnecessary. If we did want to add a new field, especially a boolean, we could do it like this:

```js
const missingData = !areaData.belonging_rate_2018 && areaData.belonging_rate_2018 !== 0;
...
return {
  chartData: data,
  areaData,
  missingData,
}
```

app/components/questions/BelongingRates.js
```diff
@@ -180,12 +208,12 @@ export default function BelongingRates(props) {
...
       </div>
       <div className="center">
-        <ExamplePieChart data={chartData} />
+        <ExamplePieChart data={chartData} dataFlag={dataFlag} />
         <br />
       </div>
       <div className="center medium-width">
         <p>
-        <span>Based on data from 2018, {areaData.belongingRateFormatted} of people in {areaData.name} agree or strongly agree that they feel a sense of belonging in their community.</span>
+        <GenerateMessage areaData={areaData} dataFlag={dataFlag} />
```
vingkan, line 216: Would be nice to have a more descriptive component name, like `AreaBelongingMessage`.

vingkan (changes requested):
Nice job learning more JavaScript and trying a React sub-component. Requested some changes for readability and maintainability.

If we let the message component be responsible for handling missing data, then we won't have to pass `dataFlag` around in so many places, which will make the whole component easier to follow, understand, and modify/add to.

app/components/questions/BelongingRates.js
fabrego524: If we check for the `missingData` in the component when it renders, would I have to pass `areaData` to the piechart as well? That was the only issue I had, considering the piechart got `data` but our area selector is passed `areaData`, which have their own structures

fabrego524: I think I've solved my issue, I wasn't aware that `areaData.rate_of_belonging_2018` and `data.value` were essentially equivalent

vingkan: Ah yes, I forgot about the pie chart. It looks like what you wrote should work, but the website build fails. I'll open another thread for that.

app/components/questions/BelongingRates.js
```diff
@@ -61,6 +60,9 @@ function transformData(response, error, selectedArea) {
 function ExamplePieChart(props) {
   const { data } = props
   
+  if(!data[0].value){
```
vingkan, line 63: The automated test for the website build is failing with this error message:

```
Error occurred prerendering page "/questions/belonging-example". Read more: https://nextjs.org/docs/messages/prerender-error
TypeError: Cannot read property 'value' of undefined
    at BelongingRates_ExamplePieChart (/home/runner/work/transithealth/transithealth/app/build/server/pages/questions.js:4075:16)
```

I think it is referring to line 63 in your change. This could be a good time to use optional chaining:

```js
if (!data?.[0]?.value) {
  return null;
}
```

Now, if `data[0]` is `undefined`, the optional chaining will evaluate early to `undefined` instead of trying to check `undefined.value` and causing an error.

Try that and let me know if it both fixes the site build and achieves your intended effect for areas that have values vs Burnside.

vingkan (changes requested):
Thanks for making those changes! The automated tests say the site doesn't build, I left a comment that should help you fix that.

app/components/questions/BelongingRates.js
fabrego524: This seems to work now, thanks for the help.

app/components/questions/BelongingRates.js
```diff
@@ -60,7 +59,11 @@ function transformData(response, error, selectedArea) {
 
 function ExamplePieChart(props) {
   const { data } = props
+  const { areaData } = props
```
vingkan, line 62: Remove this line, since areaData is not a prop, nor is it used.

vingkan (approved):
Left one comment to remove line 62, but other than that, looks good to merge! Good work fixing this case, Fabian.

PR #119, opened 2021-07-24
Improve timeline view x-axis.

### LINKED ISSUE #60: Timeline view x-axis does not change when selecting different metrics.
This is most annoying when switching from weekly to yearly metrics, because you only end up seeing the latest years.

vingkan:
Fixes #60 

- Changes XAxis from date type to number type, which is better supported by Recharts and allows dataMin and dataMax to work automatically, which dynamically resizes the x-axis domain based on the data.
- Handles the two cases where metrics that have not been selected can be present in the data:
  - Hidden via the selector
  - Metric was removed via the selector, but since we only removed a metric, we do not make another request to the API
- Changes x-axis ticks to use a format with a short month and numeric year (`Jan 2021`)
- Tries to show a tick every time a new year starts, though Recharts sometimes hides ticks to preserve gaps between ticks
- Makes the metric select element responsive on narrow screens

### DIFF
--- a/app/components/TimelineExplorer.js
+++ b/app/components/TimelineExplorer.js
@@ -27,11 +27,27 @@ const defaultColors = [
   "#a453f5",
   "#f5cd53",
 ];
-const dateFormat = new Intl.DateTimeFormat("en-US", {
+const dateFormatToolTip = new Intl.DateTimeFormat("en-US", {
   month: "long",
   day: "numeric",
   year: "numeric",
 });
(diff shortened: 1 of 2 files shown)


fabrego524 (approved):
Went ahead and tested the changes directly, everything looks good and less cluttered now, and the timeline expands more in an attempt to fit the entire line.

LGTM

vingkan:
Thanks for the thorough testing @fabrego524 !

PR #123, opened 2021-07-25
implementing logic with unit tests

alkatoutl:


api/metrics/taxitrips.py
```diff
@@ -0,0 +1,35 @@
...
+        """
+        cur = self.connection.cursor()
+        cur.execute(query)
+        rows = rows_to_dicts(cur, cur.fetchall())
+        return rows
+        
+#takes in pickup_community_area and returns most common dropoff_community_area
+    def maxdropoff(self, pickup_community_area):
+        query = """
+        SELECT max(dropoff_community_area)
```
vingkan, line 27: The max() function will return the largest value, not the most common value.

Happy to pair program with you on how to get the most common value, or you can Google something like this:

"SQL get most common value in group by"

This would be a good metric to have for a data question, but it's not the kind of metric we can show on the timeline or community scatter view. It would be great to have a metric that is total number of trips by pickup area and another by dropoff area. If you return two columns: one called area_number and one called value, that metric will plug right into our community scatter plot. You could also do a total trips per day metric, which should have one column called date and one called value to plug into the timeline view.

Let me know if anything I said above is unclear, I can show you how the metrics plug into the frontend.

api/metrics/taxitrips_tests.py
```diff
@@ -0,0 +1,95 @@
...
+            "dropoff_community_area": 23
+        }
+    ]
+    
+    connection, cur = create_test_db(
+        scripts=[
+            "./pipeline/load/taxitrip.sql"
+        ],
+        tables={
+            "dropoff": dropoff_table
```
vingkan, line 82: The table name should match the name in the schema: taxitrips

api/metrics/taxitrips.py
```diff
@@ -0,0 +1,35 @@
+from api.utils.database import rows_to_dicts
+
+#multiple metrics for the taxi trip dataset
+class TaxiTripMetrics:
+    def __init__(self, connection):
+        self.connection = connection
+        
+        
+#gets the average speed of a taxi 
+    def avgtaxispeed(self):
```
vingkan, line 10: This is a cool idea for a metric! Can you make it a group by pickup or dropoff community area and then rename the avg speed column as value? Then we could plug this into the community scatter view and see average speeds by community of pickup or dropoff.

api/metrics/taxitrips.py
```diff
@@ -0,0 +1,35 @@
...
+        query = """
+        SELECT
+            trip_miles,
+            trip_minutes/60 as trip_hours,
+            trip_miles/trip_hours as trip_speed
```
vingkan, line 15: I think this will throw an error where SQL the column trip_hours is not defined, because it gets defined at the same time as where it is being used. But maybe I am wrong about that.

vingkan (changes requested):
These are great ideas for metrics, Leilah! We need to make some change so that the output will be compatible with what the frontend views are expecting. Let me know in the #help channel if you have any questions or clarifications.

api/metrics/taxitrips.py
alkatoutl: Yeah so it does throw an error because it's defined at the same time as it's being used. I could add a new column to the dataset for trip_hours but would I have to do the transforming and loading steps again?

vingkan: No need to add it as a column in the offline pipeline, there are two ways you can do it in the query:

1. Repeat the expression:

```sql
SELECT
  trip_minutes / 60 as trip_hours,
  trip_miles / (trip_minutes / 60) as avg_speed
FROM taxitrips
...
```

2. In a subquery:

```sql
SELECT
  trip_hours,
  trip_miles / trip_hours as avg_speed
FROM (
  SELECT
    *,
    trip_minutes / 60 as trip_hours
  FROM taxitrips
)
...
```

Approach #1 should work for this case.

api/metrics/taxitrips.py
alkatoutl: Should I still group it by taxi_id, too?

vingkan: That's a good point, I think you actually want to get average speed grouped by trip_id, not by taxi_id, and then group that by community area.

You want to take the **average average** speed of trips in each community area.

alkatoutl: I was thinking of making that another metric, average trip speed, because I wanted this one to be the average speed of different taxis. Unless you think average speed of taxis isn't really needed?

vingkan: That's a good idea, make them as separate metrics:

1. avg_speed_per_pickup_area()
2. avg_speed_per_dropoff_area()
3. avg_speed_per_taxi()

The first two can plug in right away to the community scatter view, the third can be used in your data question.

All three of these are actually averages of averages, because each taxi_id can have multiple trips and each pickup or dropoff area has multiple trips. So you may actually want to use a subquery. Here are two seemingly similar queries that would have different results:

```sql
SELECT
  SUM(trip_miles) / SUM(trip_hours)
FROM taxitrips
```

```sql
SELECT
  AVG(avg_speed)
FROM (
  SELECT
    trip_miles / trip_hours AS avg_speed
)
```

I think the second version (with the subquery) is the more correct metric for "average speed". The first query would be a "macro average" (all miles divided by all hours) and the second query would be a "micro average" (average of averages).

vingkan (approved):
Great work!

Can you change the column names of your SQL queries for average speed by area to be `area` and `value`? That way they will plug into the community scatter view.

You can either do that change in this PR or refactor it in your next PR when you add those metrics to the app.

PR #124, opened 2021-07-26
added metrics to community/scatter view

alkatoutl:
I'm dying but it has been done hopefully

### DIFF
--- a/api/endpoints/community.py
+++ b/api/endpoints/community.py
@@ -1,6 +1,7 @@
 from flask import Blueprint, jsonify, request
 from api.metrics.community import CommunityMetrics
 from api.metrics.rent_burdened import RentBurdenedMetrics
+from api.metrics.taxitrips import TaxiTripMetrics
 
 
 def make_blueprint(con):
@@ -12,6 +13,7 @@ def make_blueprint(con):
     
     metric = CommunityMetrics(con)
     metric_rbu = RentBurdenedMetrics(con)
+    metric_tt = TaxiTripMetrics(con)
 
     supported_metrics = {
         "rideshare_pickups_covid": metric.rideshare_total_pickups,
@@ -36,7 +38,9 @@ def make_blueprint(con):
         "rent_burdened_2019": lambda: metric_rbu.rent_burdened(year=2019, segment="all"),
         "rent_max": lambda: metric_rbu.rent_max_burdened(),
         "rent_min": lambda: metric_rbu.rent_min_burdened(),
-        "rent_average": lambda: metric_rbu.rent_average_burden_area()
+        "rent_average": lambda: metric_rbu.rent_average_burden_area(),
+        "avg_speed_per_dropoff": lambda: metric_tt.get_avg_speed_per_dropoff(),
+        "avg_speed_per_pickup": lambda: metric_tt.get_avg_speed_per_pickup()
     }
 
 
--- a/api/endpoints/taxitrips.py
+++ b/api/endpoints/taxitrips.py
@@ -1,0 +1,24 @@
+from flask import Blueprint, jsonify, request
+from api.metrics.taxitrips import TaxiTripMetrics
+
+#creates blueprint for endpoints related to taxi trips
+def make_blueprint(connection):
+    app = Blueprint('taxitrips', __name__)
+    metric = TaxiTripMetrics(connection)
+    
+    supported_metrics = {
+        "avg_speed_per_dropoff": lambda: metric.get_avg_speed_per_dropoff(),
+        "avg_speed_per_pickup": lambda: metric.get_avg_speed_per_pickup()
+    }
+    
+    @app.route("/taxitrips/avg_pickup_speed")
+    #returns average speed by pickup area
+    def avg_pickup_speed():
+        return jsonify(metric.get_avg_speed_per_pickup())
+        
+    @app.route("/taxitrips/avg_dropoff_speed")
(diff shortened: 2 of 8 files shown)


api/endpoints/taxitrips.py
```diff
@@ -0,0 +1,24 @@
+from flask import Blueprint, jsonify, request
+from api.metrics.taxitrips import TaxiTripMetrics
+
+#creates blueprint for endpoints related to taxi trips
+def make_blueprint(connection):
```
vingkan, line 5: Making a new blueprint wasn't necessary, what you added to `api/endpoints/community.py` was all that was needed to add those metrics to the API.

Sorry if I caused confusion by saying to look at rideshare as an example. I should have said the rideshare metric in the community and timeline endpoints, not the rideshare endpoint/blueprint.

For your other custom metrics for your question, we have a file called `api/endpoints/question.py` that you can use. Or since you made this new blueprint, we can use this, that works too!

There are some other changes to make here, like removing the support_metrics map in this file (it's not being used by the code) and registering this blueprint with the server (this happens in `api/server.py`, but those can be done later when you work on your question.

api/metrics/taxitrips.py
```diff
@@ -6,9 +6,9 @@ def __init__(self, connection):
         self.connection = connection
         
     #returns avergae trip speed per pickup area    
-    def avg_speed_per_pickup(self):
+    def get_avg_speed_per_pickup(self):
         query = """
-        SELECT pickup_community_area, AVG(trip_miles/(trip_minutes/60)) as avg_speed
+        SELECT pickup_community_area as area_number, AVG(trip_miles/(trip_minutes/60)) as value
```
vingkan, line 11: Code style preference: In the future, I would recommend putting each selected expression on its own line to make the query easier to read.

api/metrics/taxitrips_test.py
```diff
@@ -88,52 +88,11 @@ def test_avg_speed_per_dropoff():
     
     metric = TaxiTripMetrics(connection)
 
-    assert metric.avg_speed_per_dropoff() == [
-        { "dropoff_community_area": 8, "avg_speed": 12.0},
-        { "dropoff_community_area": 24, "avg_speed": 12.0},
-        { "dropoff_community_area": 67, "avg_speed": 16.0}
+    assert metric.get_avg_speed_per_dropoff() == [
```
vingkan, line 91: Good job refactoring the test.

api/questions/taxitrips.py
```diff
@@ -0,0 +1,45 @@
...
+        ORDER BY COUNT(*) desc
+        LIMIT 1
+        """.format(pickup_community_area = pickup_community_area)
+        cur = self.connection.cursor()
+        cur.execute(query)
+        rows = rows_to_dicts(cur, cur.fetchall())
+        return rows
+        
+    #gets most used payment type by pickup location
+    def get_payment_type_by_pickup(self):
```
vingkan, line 19: Nice, another interesting idea for a metric! We can review this more when you start using it, since I don't want to block launching your metric.

pipeline/extract/taxi_trips.sql
```diff
@@ -1,5 +1,5 @@
 SELECT date_trunc_ymd(trip_start_timestamp) as ymd, 
-    trip_id, 
+    trip_id ,
```
vingkan, line 2: You could remove this space.

api/endpoints/community.py
```diff
@@ -36,7 +38,9 @@ def make_blueprint(con):
         "rent_burdened_2019": lambda: metric_rbu.rent_burdened(year=2019, segment="all"),
         "rent_max": lambda: metric_rbu.rent_max_burdened(),
         "rent_min": lambda: metric_rbu.rent_min_burdened(),
-        "rent_average": lambda: metric_rbu.rent_average_burden_area()
+        "rent_average": lambda: metric_rbu.rent_average_burden_area(),
+        "avg_speed_per_dropoff": lambda: metric_tt.get_avg_speed_per_dropoff(),
+        "avg_speed_per_pickup": lambda: metric_tt.get_avg_speed_per_pickup()
```
vingkan, line 43: This works, but technically this doesn't need to be a lambda function.

The reason we use lambda functions in the supported_metrics map is that sometimes, we want to modify a query with a variable, but we don't want to write an entirely new method. However, supported_metrics expects a method with no parameters. So we use a lambda function to create a method that has no parameters by "wrapping" our call to the method with the parameter we want. This is called "currying."

However, your avg_speed methods do not take any parameters, so you could list them like this, with no parentheses after the method, which passes the method as a variable, rather than calling it:

```python
{
        ...
        "avg_speed_per_dropoff": metric_tt.get_avg_speed_per_dropoff,
        "avg_speed_per_pickup": metric_tt.get_avg_speed_per_pickup
}
```

No need to change it though, I noticed just now that the same thing was done for the rent burdened metrics.

vingkan (approved):
Great work, Leilah! This is good to merge, which will deploy your metrics to the production website. Congratulations, that's a big milestone!

I left some other optional comments, you don't have to change them before you merge, I just added them for your information. Let me know if you have any questions about them.

divyashie (approved):
Except for the few comments that Vinesh has, this is great work! Awesome Leilah

vingkan:
Hey Leilah, a bunch of new PRs have been merged since we approved your PR, you should pull the latest from main, merge that into your branch, and resolve conflicts before merging this PR.

PR #125, opened 2021-07-26
Added new page detailing the new cybersecurity task

olakukielko:


vingkan (approved):
LGTM!

docs/pages/headeraudit.md
```diff
@@ -0,0 +1,24 @@
...
+It gives a letter grade for a URL, as well as gives details of what can be fixed and what can be improved.
+
+## Part 1 - Report
+Analyze the security standards implemented on our [project website](https://scarletstudio.github.io/transithealth/) using the free header checking tool [SecurityHeaders.com](https://securityheaders.com).  
+
+- Describe each header and how it applies to the usage of our site.  It’s possible some headers are not relevant to our site.  Ask if you are not sure.
+- Determine which headers are missing
+- Determine which headers are present, but could be improved (if any)
+- Determine which headers are properly implemented.
+- Assess & describe your findings in a report that is written in a jupyter notebook.
```
vingkan, line 15: Do you think the report should be in Jupyter or in a Google Doc? I think a doc might be easier to leave feedback on, and we don't necessarily need to run code in a notebook as part of the report, perhaps separately.

docs/pages/headeraudit.md
```diff
@@ -0,0 +1,24 @@
...
+
+There are many free tools available that can help us audit the security standards of a given site. 
+A very common, yet overlooked security standard is the HTTP Headers that a site transfers when a user accesses their site.
+[SecurityHeaders.com](https://securityheaders.com/) is one great free tool that we can use. 
+It gives a letter grade for a URL, as well as gives details of what can be fixed and what can be improved.
+
+## Part 1 - Report
+Analyze the security standards implemented on our [project website](https://scarletstudio.github.io/transithealth/) using the free header checking tool [SecurityHeaders.com](https://securityheaders.com).  
+
+- Describe each header and how it applies to the usage of our site.  It’s possible some headers are not relevant to our site.  Ask if you are not sure.
```
vingkan, line 11: Should we have a first task be just a single specific header that we know is relevant?

olakukielko: @vingkan  i think it's important to view the report as a whole & see everything missing, everything that could be improved, and everything that we are doing right.  otherwise, the grade doesn't make sense.  

we could modify Task 1, Bulletpoint 1 (describe each header and how it applies to the usage of our site) so that you only have to describe one missing header, and update task 2 so that the fix applies to that header. i'll send an update later today.

PR #126, opened 2021-07-27
Sidewalk Cafe Metrics

asudeozturk:
Create  'number of sidewalk cafe permits per day' and 'number of sidewalk cafe permits per year' metrics with units tests


api/metrics/sidewalk_cafe.py
```diff
@@ -0,0 +1,43 @@
...
+        self.con = con
+    
+    def get_total_permits_day(self):
+        """
+        Returns the number of permits issued per day.
+        """
+        query = """
+        SELECT
+            issued_date_dt as date,
+            count(issued_date_dt) as number_of_permits
```
vingkan, line 18: Rename this column to `value` so that it automatically plugs into the timeline view.

vingkan (approved):
Great work, Asude! You wrote great unit tests all on your own.

We also uncovered bug #127 which is unrelated to this change. Great find!

Issue #128, opened 2021-07-27
Helper method to extract dates fails if column has nulls


vingkan:
## To Reproduce
Steps to reproduce the behavior.

vingkan:
Thank you @asudeozturk for catching this!

PR #129, opened 2021-07-27
Extracted all of the remaining sidewalk cafe permits

asudeozturk:
Extract all sidewalk cafe permit data

### DIFF
--- a/api/metrics/sidewalk_cafe_test.py
+++ b/api/metrics/sidewalk_cafe_test.py
@@ -61,9 +61,9 @@ def test_total_permits_per_day():
         ], "Should have 4 results. Check if number of permits are correct"
     
     assert metric.get_total_permits_year() == [
-        {"year": "2002-01-01", "value": 3 },
-        {"year": "2005-01-01", "value": 1 },
-        {"year": "2021-01-01", "value": 1 }
+        {"date": "2002-01-01", "value": 3 },
+        {"date": "2005-01-01", "value": 1 },
+        {"date": "2021-01-01", "value": 1 }
         ], "Should have 3 results. Check if number of permits are correct"
     
 
--- a/pipeline/extract/sidewalk_cafe.sql
+++ b/pipeline/extract/sidewalk_cafe.sql
@@ -10,3 +10,4 @@ SELECT
     latitude,
     longitude
 WHERE issued_date <= "2021-07-15"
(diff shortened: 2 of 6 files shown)


api/metrics/sidewalk_cafe_test.py
```diff
@@ -61,9 +61,9 @@ def test_total_permits_per_day():
         ], "Should have 4 results. Check if number of permits are correct"
     
     assert metric.get_total_permits_year() == [
-        {"year": "2002-01-01", "value": 3 },
-        {"year": "2005-01-01", "value": 1 },
-        {"year": "2021-01-01", "value": 1 }
+        {"date": "2002-01-01", "value": 3 },
```
vingkan, line 64: Great catch! I missed this on the last review.

pipeline/transform/sidewalk_cafe.py
```diff
@@ -25,7 +25,8 @@
 
 #Parse dates
 df = extract_data_portal_dates(raw_df, col="issued_date")
-df = extract_data_portal_dates(df, col="expiration_date")
+#If expiration date is null this will fail
```
vingkan, line 28: Another good catch! Opened bug #128 to fix this.

pipeline/tests/database_test.py
```diff
@@ -28,6 +28,7 @@ def test_table_counts():
     test.assert_table_count("rideshare", 671938)
     test.assert_table_count("cta_train_ridership", 104564)
     test.assert_table_count("taxitrips", 100000)
+    test.assert_table_count("sidewalk_cafe", 19858)
```
vingkan, line 31: Thanks for adding this, this test will help others know if they have the correct version of the database.

pipeline/extract/sidewalk_cafe.sql
```diff
@@ -10,3 +10,4 @@ SELECT
     latitude,
     longitude
 WHERE issued_date <= "2021-07-15"
+LIMIT 100000
```
vingkan, line 13: This was a tricky point about the data portal that was not documented: If the query has no limit, by default it will return up to 1000 rows. If you have more than that, you need to explicitly set a very large limit.

vingkan (approved):
Great work, Asude! We found a lot of interesting issues that others will probably run into And the data looks very interesting!

PR #130, opened 2021-07-27
Ada cta stops

MariaGarciaCS:
- Added transformation to include community area number
- Wrote queries in notebook to compare cta train ridership at ada and non ada stations


vingkan (approved):
Great work Maria!

PR #131, opened 2021-07-27
Update load

talemu:


vingkan (approved):
LGTM!

pipeline/Makefile
```diff
@@ -47,16 +48,16 @@ data/loaded/all_tables.txt:
...
+	make data/loaded/disabilities.txt
 	make data/loaded/population.txt
 	make data/loaded/rent_burdened_households.txt
 	make data/loaded/income.txt
 	make data/loaded/covid_spread.txt
 	make data/loaded/traffic_intensity.txt
+	make data/loaded/escooter.txt
 	make data/loaded/rideshare.txt
-	make data/loaded/disabilities.txt
-	make data/loaded/cta_train_stops.txt
```
vingkan, line 57: Oops, my bad for missing this in my review. We accidentally deleted the CTA train stops load step from the database load step. @MariaGarciaCS we'll have to add this back, re-run the pipeline, and push new compressed database files to main.

We should add a database test for CTA train stops so there will be an automatic failure if it gets automatically removed.

Issue #99, opened 2021-07-22
Improve the metrics selector


vingkan:
We have a common component called `MetricSelector`, located in [`app/components/Common.js`](https://github.com/scarletstudio/transithealth/blob/6d9ed46bb396ae6d4a7b07f7e01e97778684be42/app/components/Common.js#L1), that is currently used by the scatter view and timeline view to allow website visitors to choose metrics.

Screenshot of this component in action, being used to selection a y-axis metric on the scatter view:

<img width="1388" alt="Screen Shot 2021-07-21 at 8 32 49 PM" src="https://user-images.githubusercontent.com/11896652/126579235-e3b15a6e-0068-46d1-a176-30cb321a26e4.png">

It could be a good stretch project to improve this reusable component:

- When clicked, pop up a modal for choosing a metric
- Make the list of metrics searchable
- Group the metrics, maybe by topic or dataset
- Add a setting or create a new component to select multiple metrics

PR #136, opened 2021-07-28
Implemented Search for Scatter modal

fabrego524:
Added a new input search bar within the new modal for the Scatter explorer page, as mentioned in #99. Uses live results rather than needing to press search. Also dynamically displays the current search results based on the input text in the search bar.


fabrego524:
I quickly added the new modal functionality to the Timeline Explorer page as well. The styling seems to fit fine, if not a tiny bit roughly.

vingkan (approved):
Awesome! This will be very valuable, especially as we add more and more metrics.

PR #137, opened 2021-07-29
Add Contentsecuritypolicy header

Jmacias2019:
Fixed Style Policy


vingkan (approved):
LGTM! If this doesn't work in production, let's roll it back. If it works we celebrate!

PR #138, opened 2021-07-29
added questions with tests and endpoint, sidewalk test fails?

alkatoutl:


alkatoutl:
when I run tests in my cloud 9, all tests pass except test_table_counts where `Table sidewalk_cafe should have 19858 rows, not 1000 rows.` but the tests here on github is saying all my question tests fail?

vingkan:
Merge main into your branch and run `run pipeline-quick` to see if that gets the tests to pass?

vingkan:
I think the queries are incorrect. I am not 100% sure about this, but I think if you use a group by without an aggregate like you are doing in your outermost query, the result could be any of those rows, leading to unpredictable behavior on different systems.

I assume you are trying to get the most common payment type for each area. I don't think you need three nested queries to do that. One way to do that query where the output is deterministic would be like this:

```sql
SELECT area, payment
FROM (
  SELECT
    area,
    payment,
    count(*) as count
  FROM taxitrips
  GROUP BY area, payment
)
WHERE count == MAX(count)
GROUP BY area
```

alkatoutl:
I'm testing the tested data table and my query on an outside SQL practice website and it works there, too. But I'll try to use your query!

vingkan:
The 19000 number is the correct number of rows, Asude fixed an error in the sidewalk cafes extraction that was causing it to get only the first 1000. So of you reran the pipeline you should pass that test. Double check that the test is expecting the right number, not 1000.

alkatoutl:
The sidewalks test passes now on my end. When I use your query my test fails now, though. Should I still commit it?

divyashie (approved):
Great work!

PR #139, opened 2021-07-29
Added choices for years in rent burdened households question

wpj0110:


app/components/questions/RentBurdenRates.js
```diff
@@ -21,34 +21,39 @@ const EMPTY_AREA = {
...
-function transformData(response, error, selectedArea) {
+function transformData(response, error, selectedArea, selectedYear) {
   if (!response || error) {
     return {
       chartData: [],
       areaData: EMPTY_AREA,
     }
   }
+  
+  var rentYear = ((selectedYear === 2019) ? 'rent_burdened_2019'
```
vingkan, line 32: You could make the selector data have the metric name as options to avoid this check. For example:

```js
function YearSelector(props) {
  ...
  <select>
    {props.data.map((d, i) => (
      <option value={d.metric}>{d.year}</option>
    ))}
  </select>
  ...
}
...
const years = [
  { year: 2018, metric: 'rent_burdened_2018' },
  ...
]
...
<YearSelector data={years} ... />
```

vingkan (approved):
Looks good! Great job adding another selector, William. Can you add a screenshot of how the page looks to your PR as documentation of what this change does?

app/components/questions/RentBurdenRates.js
```diff
@@ -199,6 +229,7 @@ export default function RentBurdenedRates(props) {
       <div className="center medium-width">
         <h2>Rent Burdened Households Rate by Area</h2>
         <AreaSelector data={data} handleChange={setSelectedArea} />
+        <YearSelector data={years} handleChange={setSelectedYear} />
```
vingkan, line 232: Do these selectors show up on the same or different lines? Could use some styling. Would be great to see a screenshot with the PR to know how it looks.

wpj0110: temp
I think I will change it a bit.

app/components/questions/RentBurdenRates.js
wpj0110: Uhm... yeah I think I'll just keep it XD
True. It is awkward regardless.

wpj0110:
temp2
Made it a little more compact when selecting area and year.

PR #141, opened 2021-07-29
Remove 2000 Total Population Metric

fabrego524:
During some testing of the new modal's UI on a mobile platform, I noticed that choosing the `2000 Total Population` metric on the app resulted in an empty scatter plot and a fully opaque map. Logging the value of the data in the code showed that no Population data was being returned. This can be seen below

image

It was a bit unclear why this was the case, so I did some further testing by pulling the latest for main and running `run pipeline-quick`. After a quick check of the database using SQLite, using the query 
```SQL
SELECT MAX(value), period_end_year
FROM population
GROUP By period_end_year
```

It returned data for years from 2010 to 2019.

image

I also went ahead and double-checked the Chicago Health Atlas API, and it appears the data for the year 2000 was removed at some point.

I asked @vingkan about the best solution and we've decided to simply remove this particular metric. The 2010 and 2019 Total Population metrics remain unchanged and work as intended.


vingkan (approved):
Great work noticing this and finding the root cause, Fabian! This was an interesting issue out of our control, as the Chicago Health Atlas API can change without warning. This is a pragmatic fix, as luckily, we can do without the 2000-2009 data.

PR #145, opened 2021-07-31
Search for sidewalk permits

vingkan:
Added a new question that lets visitors search for sidewalk permits by restaurant name.

Does this new feature have any security vulnerabilities?


vingkan:
Bad things happen when you merge without getting your code reviewed.

PR #148, opened 2021-07-31
Modal Search Result Grouping

fabrego524:
Implemented a new component that returns the modal's search results grouped under a corresponding tag or dataset. Groups are returned by the number of metrics grouped under them, from greatest to least. This also adds two new fields for metrics on the frontend: Dataset and Description. 

Dataset refers to the data that is extracted, transformed, and loaded into the database, while Description goes into detail about what the metric is showing.

The Modal Before Grouping: 

image

The Modal After Grouping:

image

Currently, the rendering component works. However, upon clicking one of the metrics, a TypeError occurs. I'll be sure to do some debugging when possible in order to track down the cause of this, since the original


app/components/SearchableMetricSelector.js
```diff
@@ -1,25 +1,62 @@
...
+      var tempArr = sortedTagsArr[i]
+      var tempDict = {};
+      tempDict[tempArr[0]] = tempArr[1]
+      retArr.push(tempDict)
+    }
+    return retArr
+  }
+  
+function CreateTagDictionary(supportedMetrics){
+  const groups = [...new Set(Object.keys(supportedMetrics).map(q => supportedMetrics[q].dataset))];
```
vingkan, line 16: I think what you are trying to do here is get a list of the unique group IDs. Write that in a comment above to help make this line more readable.

app/components/SearchableMetricSelector.js
```diff
@@ -1,25 +1,62 @@
...
+
 function Modal(props){
   const { supportedMetrics, onClose, selectMetric, show } = props;
   const [searchText, setSearchText] = useState("")
   if(!show){
     return null;
   }
+  var tagDictionary = CreateTagDictionary(supportedMetrics);
+  var tagDictArray = CreateDictArray(tagDictionary);
+  tagDictArray = SortDictionary(tagDictionary)
```
vingkan, line 51: Style nitpick: we usually save uppercase first letters for names of Components, since this is a helper function, it can start with lowercase.

app/components/SearchableMetricSelector.js
```diff
@@ -38,17 +75,14 @@ function Modal(props){
...
-                >{supportedMetrics[k].name}</p>
-              ))}
-            </div>
+            <GroupedMetricSearchResults
+              searchText={searchText} 
+              supportedMetrics={supportedMetrics} 
+              onClose={onClose}
+              selectMetric={selectMetric}
+              tagDictArray={tagDictArray}
+              tagDictionary={tagDictionary}
```
vingkan, line 84: Could you move the computation of `tagDictArray` and `tagDictionary` into `GroupedMetricSearchResults`? I don't think it needs to be computed in `Modal`, so it should be an implementation detail of `GroupedMetricSearchResults`.

This will help declutter the `Modal` state variables as well as make `GroupedMetricSearchResults` a true drop-in replacement for `SimpleMetricSearchResults`.

app/components/SearchableMetricSelector.js
```diff
@@ -58,11 +92,109 @@ function Modal(props){
...
+function GroupedMetricSearchResults(props){
+  const { searchText, supportedMetrics, onClose, selectMetric, tagDictionary, tagDictArray } = props
+  const searchTextLower = searchText.toLowerCase().trim()
+  const searchResults = searchTextLower.length === 0 
+    ? Object.keys(supportedMetrics)
+    : Object.keys(supportedMetrics)
+      .filter((k) => {
+        const metric = supportedMetrics[k]
+        const metricNameLower = metric.name.toLowerCase()
+        return metricNameLower.indexOf(searchTextLower) > -1
```
vingkan, line 139: Are you also going to consider the dataset name in the search filter? It's okay if not, that can be done later or not done at all.

We may also want to include metric description in the search.

Either way, it's possible that for both `SimpleMetricSearchResults` and `GroupedMetricSearchResults`, the search filter function has exactly the same logic. In that case, you can reorganize it into a common helper function and then call that function in each component like this:

```js
const searchResults = filterMetricsForSearch(supportedMetrics, searchText);
```

app/components/SearchableMetricSelector.js
```diff
@@ -58,11 +92,109 @@ function Modal(props){
...
+      })
+
+  var elements = [];
+  for (var i = 0; i < tagDictArray.length; i++){
+    for (var key in tagDictArray[i]){
+      const metricArr = tagDictionary[key]
+      var count = 0;
+      for (var j = 0; j < searchResults.length; j++){
+        if (supportedMetrics[searchResults[j]].dataset.indexOf(key) > -1 && count < 1){
```
vingkan, line 148: Why do you have `count < 1`?

app/site/metrics.js
```diff
@@ -22,144 +22,192 @@ export const communityMetrics = {
...
+    dataset: "Population",
+    description: "Average population over the time period that ends on 2010.",
     format: Formatter.numberInThousands,
     fullFormat: Formatter.numberWithCommas,
   },
   median_income_2010: {
     name: "2010 Median Income",
     units: "dollars",
+    dataset: "Income",
+    description: "Income in the past 12 months for 2010, in inflation-adjusted 2017 dollars.",
```
vingkan, line 82: Thanks for adding these! I can fill in the others. We'll also want to add `source`, but that can be done in another PR.

app/styles/app.css
```diff
@@ -248,17 +248,38 @@ input {
...
 }
 
+.metricResult .metricMetaData p {
+    margin-top: 2px;
+    margin-bottom: 2px;
+}
+
+.searchResults .metricMetaData {
+    font-size: 0.8em;
+    padding-left: 20px;
```
vingkan, line 280: For this and the above, try to use units of `em` instead of `px` so that the design scales correctly if the user changes their default browser font size.

app/styles/app.css
```diff
@@ -248,17 +248,38 @@ input {
...
 }
 
-.searchableMetricSelector .searchResults p {
+.searchableMetricSelector .metricResult {
     border-bottom: 1px solid #D3D3D3;
 }
 
-.searchableMetricSelector .searchResults p:hover {
+.searchableMetricSelector .metricResult:hover{
+    background-color: rgba(180, 180, 180, 0.5);
```
vingkan, line 265: You can make a CSS variable (`var`) for this. Cloud9 complains, but look at the synntnax used in other places in the file, that's valid.

vingkan (changes requested):
Excellent start, Fabian. Good decomposition of the search results and great work learning a lot of new JavaScript. I left some comments, also make sure you:

- Use `let` and `const` like we talked about to avoid scoping errors
- Fix the React validate DOM warnings for <div> in <p> and <p> in <p>

app/components/SearchableMetricSelector.js
```diff
@@ -1,25 +1,62 @@
...
+  }
+  
+  return tagDictionary
+}
+
+function CreateDictArray(tagDictionary){
+  var tagDictArray = [];
+  var chunk_size = 1;
+  for ( var cols = Object.entries( tagDictionary ); cols.length; ){
+    tagDictArray.push( cols.splice(0, chunk_size).reduce( (o,[k,v])=>(o[k]=v,o), {}));
```
vingkan, line 38: Can you rewrite this to make it more readable? It also looks like it's copied from somewhere, so either add a link to the source or rewrite it in a way that you understand and think will be readable to others on our team.

app/components/SearchableMetricSelector.js
fabrego524: This `count < 1` dictates that each group is only printed once. I'm sure there is a better way of doing this, previously it would print the rideshare group once for every rideshare metric in the dictionary

app/components/SearchableMetricSelector.js
```diff
@@ -1,25 +1,75 @@
 import { useEffect, useState } from 'react'
 
-function Modal(props){
-  const { supportedMetrics, onClose, selectMetric, show } = props;
-  const [searchText, setSearchText] = useState("")
-  if(!show){
-    return null;
+function sortDictionary(dictionary){
+    var sortedTagsArr = Object.entries(dictionary).sort((a,b) => b[1].length - a[1].length)
```
vingkan, line 4: Add a comment that Object.entries gives each element as an array of [key, value], so you use b[1] to get the value, which is the array of metric objects, then sort based on number of matching metrics in each group.

app/components/SearchableMetricSelector.js
```diff
@@ -1,25 +1,75 @@
...
+function createTagDictionary(supportedMetrics){
+  const groups = [...new Set(Object.keys(supportedMetrics).map(q => supportedMetrics[q].dataset))];
+  const metricRecords = Object.keys(supportedMetrics)
+  var tagDictionary = {};
+  
+  for (var i in groups){
+    var groupedRecords = [];
+    for(var j in Object.keys(supportedMetrics)){
+      var record = Object.keys(supportedMetrics)[j]
+      if(groups[i].indexOf(supportedMetrics[record].dataset) > -1 ){
```
vingkan, line 25: This should be `groups[i] === supportedMetrics[i]. dataset` since they should be exactly equal in order to belong to the group. You use the indexOf if you only want a partial match, such as in a search string.

app/components/SearchableMetricSelector.js
```diff
@@ -1,25 +1,75 @@
...
+      var tempArr = sortedTagsArr[i]
+      var tempDict = {};
+      tempDict[tempArr[0]] = tempArr[1]
+      retArr.push(tempDict)
+    }
+    return retArr
+  }
+  
+//Create a dictionary of all unique tags/datasets that exist, each key holding an array of metrics belonging to that tag/dataset
+function createTagDictionary(supportedMetrics){
```
vingkan, line 16: Later we can revise this method to be more runtime efficient. I left a comment below about correctness.

app/components/SearchableMetricSelector.js
vingkan: @fabrego524 Did you see this comment?

app/components/SearchableMetricSelector.js
vingkan: Ah I see. Thanks for the explanation. Can you add a comment to explain that? We can refactor it in the future.

app/styles/app.css
vingkan: @fabrego524 did you see this comment?

vingkan (changes requested):
Replied to some comments and closed the ones you finished. Almost there!

vingkan (approved):
Great work, Fabian! We can continue to refactor this, but want to get it merged. This was a great stretch for you to practice some new JavaScript and React skills.

PR #150, opened 2021-08-01
Disabilities transportation question

MariaGarciaCS:
Finished first table. There's a couple more I want to add and a bar chart too, but this can be displayed on the site as it it rn.


app/site/questions.js
```diff
@@ -6,7 +6,11 @@ import ResidentsWithDisabilities from '../components/questions/ResidentsWithDisa
 import BelongingRates from '../components/questions/BelongingRates'
 import RentBurdenedRates from '../components/questions/RentBurdenRates'
 import SidewalkCafePermitsYears from '../components/questions/SidewalkCafePermitsYears'
+<<<<<<< HEAD
+import DisabilitiesAndTransportation from '../components/questions/DisabilitiesAndTransportation'
+=======
```
vingkan, line 11: There are some merge markers in this file, make sure you resolve the merge conflicts!

vingkan (changes requested):
Awesome! Can you fix the merge conflicts and add a screenshot of how the page looks in your PR description for documentation?

There is a lot of SQL logic to review and I don't want to block you, so if the site builds and the page works, we can approve this.

Try clicking on the column names to sort them to see whether or not that causes an error. If you can sort all the columns without errors that'll be good.

app/site/questions.js
MariaGarciaCS: image

vingkan (approved):
Excellent, Maria! This is a great foundation for your data question. Glad we could get the work you've been doing checked in.

PR #151, opened 2021-08-01
implemented metric

talemu:


vingkan (changes requested):
Good work Tabor. To make these metrics integrate with the community scatter plot, we need the columns returned to be called area_number and value. Can you change some of these metrics that take a specific community area as value x to instead do a group by and get the value for all community areas?

Let me know if this is not clear or check out the functions in api/metrics/community.py to see examples.

talemu:
[image: image.png]
Similar to this?

divyashie (approved):
Lgtm! Great work 👍🏽

api/endpoints/community.py
```diff
@@ -13,14 +12,14 @@ def make_blueprint(con):
...
     metric_rbu = RentBurdenedMetrics(con)
-    metric_tt = TaxiTripMetrics(con)
 
     supported_metrics = {
         "rideshare_pickups_covid": metric.rideshare_total_pickups,
         "rideshare_pooled_trip_rate_2018": lambda: metric.rideshare_pooled_trip_rate(year=2018),
         "rideshare_pooled_trip_rate_2019": lambda: metric.rideshare_pooled_trip_rate(year=2019),
         "rideshare_pool_request_rate_2018": lambda: metric.rideshare_pool_request_rate(year=2018),
         "rideshare_pool_request_rate_2019": lambda: metric.rideshare_pool_request_rate(year=2019),
+        "total_population_2000": lambda: metric.population(year=2000, segment="all"),
```
vingkan, line 22: Remove this line, we had to remove this metric because Chicago Health Atlas removed the data for it.

I think you started your branch in an older state where it still existed, which is why you are adding it back.

api/endpoints/community.py
```diff
@@ -13,14 +12,14 @@ def make_blueprint(con):
     
     metric = CommunityMetrics(con)
     metric_rbu = RentBurdenedMetrics(con)
-    metric_tt = TaxiTripMetrics(con)
```
vingkan, line 16: Add this back, you are accidentally deleting the Taxi metrics from this endpoint.

api/endpoints/community.py
```diff
@@ -37,9 +36,7 @@ def make_blueprint(con):
         "rent_burdened_2019": lambda: metric_rbu.rent_burdened(year=2019, segment="all"),
         "rent_max": lambda: metric_rbu.rent_max_burdened(),
         "rent_min": lambda: metric_rbu.rent_min_burdened(),
-        "rent_average": lambda: metric_rbu.rent_average_burden_area(),
-        "avg_speed_per_dropoff": lambda: metric_tt.get_avg_speed_per_dropoff(),
-        "avg_speed_per_pickup": lambda: metric_tt.get_avg_speed_per_pickup()
```
vingkan, line 42: Add these two average speed metrics back after the rent burdened, this is accidentally deleting them from the endpoint.

api/metrics/escooter_metric.py
```diff
@@ -0,0 +1,127 @@
...
+        rows = rows_to_dicts(cur, cur.fetchall())
+        return rows
+
+    def avg_distance_based_on_start_can(self):
+        """
+        Returns the avg distance traveled based on starting community area numbers
+        """
+        query = """
+        SELECT
+            start_community_area_number, avg(avg_trip_distance)
```
vingkan, line 34: Use `AS` to rename the columns to area_number and value, so that they will match the column names the community view is expecting.

api/metrics/escooter_metric.py
```diff
@@ -0,0 +1,127 @@
...
+        rows = rows_to_dicts(cur, cur.fetchall())
+        return rows
+    
+    def avg_distance_based_on_end_can(self):
+        """
+        Returns the avg distance traveled based on end community area numbers
+        """
+        query = """
+        SELECT
+            end_community_area_number, avg(avg_trip_distance)
```
vingkan, line 51: Same here, change the column names using `AS` to area_number and value.

api/metrics/escooter_metric.py
```diff
@@ -0,0 +1,127 @@
...
+        return rows
+    
+    def number_of_trips_based_on_start_cn (self):
+        """
+        Returns the number of trips from all communities that had at least
+        one ride start there
+        """
+        query = """
+        SELECT
+            start_community_area_number, sum(count_trip_id)
```
vingkan, line 85: Same comment about column names renaming.

api/metrics/escooter_metric.py
```diff
@@ -0,0 +1,127 @@
...
+        return rows
+        
+    def number_of_trips_based_on_end_cn (self):
+        """
+        Returns the number of trips from all communities that had
+        at least one ride end there
+        """
+        query = """
+        SELECT
+            end_community_area_number, sum(count_trip_id)
```
vingkan, line 103: Same comment about renaming column names.

api/metrics/escooter_metric_test.py
```diff
@@ -0,0 +1,57 @@
...
+    ], "Should have three results for this method"
+    
+    assert metric.total_escooter_rides() == [
+        { "sum(count_trip_id)" : 9}
+    ], "Not same number of total trips"
+    
+    assert metric.number_of_trips_based_on_end_cn() == [
+        {"end_community_area_number": 3, "sum(count_trip_id)": 1},
+        {"end_community_area_number": 25, "sum(count_trip_id)": 6},
+        {"end_community_area_number": 31, "sum(count_trip_id)": 2}
```
vingkan, line 55: After you finish renaming the columns in your query results you'll have to change them in the assertions here to check the new output.

vingkan (changes requested):
Thanks for updating the queries to use group bys. Two things missing before we can merge, see my comments:
1. Need to use `AS` to rename the output columns of your metric queries to be area_number and value, since that is what community scatter metrics are expecting. I commented on the places where you need to change them. You'll also have to update the output fields in your test assertions.
2. The merge conflicts in the api/endpoint/community.py were handled wrong so it is accidentally deleting the taxi metrics, please add those back as well as also delete the population metric for 2000. See my comment for more details.

talemu:
Sent another pr regarding the changes you requested. Let me know how they look.

vingkan (approved):
Good job, Tabor!

PR #152, opened 2021-08-01
Disabilities transportation question

MariaGarciaCS:
Added table 2 that looks at percent change between ADA and non ADA stops.

image


vingkan:
Uh oh, I merged your previous PR and it looks like you kept working on that branch which has caused some conflicts with your own files. Not sure how that happened.

That being said, I thought for the ADA vs non-ADA you wanted to do a bar chart? You can do both.

Maybe it would be easiest to close this pull request, pull the latest from main, create a new branch off of main, and then add this work that you were doing in this PR on that branch and open a new PR. That should make sure we don't get the merge conflicts.

vingkan (changes requested):
Your changes tab contains changes that were added in your previous PR and the merge conflicts are also with files from your previous PR, so I think something went wrong. Check out my comment: https://github.com/scarletstudio/transithealth/pull/152#issuecomment-890444275

MariaGarciaCS:
Will do!  And yes, I'll add a bar chart to the ADA non  ADA section

Issue #153, opened 2021-08-01
Searchable metrics selector group results are not ordered by number of matching metrics in group


vingkan:
## Description

When you search for metrics in the metric selector, the grouped results are not ordered by the number of matching metrics in each group.

## To Reproduce

- Go to scatter view
- Click one of the metric selector
- Search for 2019

<img width="522" alt="Screen Shot 2021-08-01 at 1 54 59 PM" src="https://user-images.githubusercontent.com/11896652/127782291-6fc59637-1ce1-44f4-b042-511173764a64.png">

## Expected Behavior

Rideshare group should come first because it has two matching metrics, while the others have just one matching metric.

PR #156, opened 2021-08-01
Corrected Modal Grouping Behavior

fabrego524:
Applies a fix in regards to #153. Previous behavior was not necessarily a bug, but was a product of when the tag dictionary was sorted. Now the dictionary and groups are sorted based on the current search text rather than all possible metrics. 

The new behavior can be seen below.

image


vingkan (approved):
Good fix, Fabian. Very clean diff.

PR #157, opened 2021-08-01
Upgrade NextJS to version 11

vingkan:


vingkan:
Not necessary for now, can consider this in the future.

PR #158, opened 2021-08-01
Fix bug where removing metrics from timeline view then adding metrics causes an error.

### LINKED ISSUE #147: Timeline view error when adding a metric after deleting all metrics
## Description
Error when adding a metric after deleting all metrics.
There are no errors for Weekly Rideshare pickups and Weekly COVID cases if you add one of them after deleting all the metrics. 
temp3
After adding the Weekly Rideshare pickups metric,
temp4
However, for all the rest, there is an error.

## To Reproduce

1. Delete all metrics in the timeline view
2. Add a metric (Note: If you add Weekly Rideshare pickups or Weekly COVID cases, there will be no error)
temp5
3. Should be the error
temp6
## Expected Behavior
Adding a metric after deleting all metrics should simply show a new timeline view of that newly added metric.

vingkan:
Fixes #147 

Very similar root cause and fix as in #103 

**Root Cause:** Timeline view tries to render as soon as the selected metrics change, but the data has not yet arrived, causing an error when it tries to format metrics it does not have yet.

**Solution:** Skip render if no selected metrics are in the data. When the response comes, the timeline chart will be able to render the new metrics.

Recording showing the fix in action:

https://www.loom.com/share/07ae403b62e14115bfd1d9c3886bb16c

### DIFF
--- a/app/components/TimelineExplorer.js
+++ b/app/components/TimelineExplorer.js
@@ -94,8 +94,17 @@ function TimelineChart({ data, metrics }) {
   const selectedMetricIds = metrics
     .filter(m => m.axis !== "none")
     .map(m => m.id);
+    
+  const dataWithSelectedMetric = data
+    .filter((d) => (
+        selectedMetricIds.filter(m => m in d).length > 0
+      ));
   
-  if (data?.length === 0 || selectedMetricIds.length === 0) return null;
+  if (data?.length === 0 ||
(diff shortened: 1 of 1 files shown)


vingkan:
@asudeozturk this is a fix for the bug you found last week where deleting metrics from the timeline chart and then trying to add new ones causes an error. We can review it together during our next call.

asudeozturk (approved):

PR #162, opened 2021-08-02
Cypress Setup

fabrego524:
Installs Cypress with the intention of implementing End-to-End testing for the project. Also creates a basic sample test that ensures cypress works correctly. 

This functionality can be seen below, where one test is meant to pass and one intentionally fails.

image


vingkan (approved):
Excellent start, Fabian!

As a note, we found out that we need Node version 12 or higher to use Cypress. I was working on that change as part of #157 and that has a risk of breaking other people's workflow, so I can take care of that.

PR #164, opened 2021-08-02
Sidewalk cafe transform area

asudeozturk:
Add metrics to show the number of sidewalk cafe permits per community area


vingkan (approved):
Great work, Asude! Glad that you were able to get the community area added to the sidewalk cafe permit data.

PR #165, opened 2021-08-02
second frontend component

alkatoutl:


app/components/questions/TaxiPaymentMethod.js
```diff
@@ -0,0 +1,115 @@
...
+import useFetch from 'use-http'
+import { Table } from '../../components/Table'
+import { Notification } from '../../components/Notification'
+import {
+  Formatter
+} from '../../site/metrics'
+
+const COMMON_PAYMENT_ENDPOINT = `${process.env.NEXT_PUBLIC_API}/question/taxitrips`;
+
+const COMMON_PPAYMENT_COLS = [
```
vingkan, line 11: Small comment on typo: "ppayment" It's not a problem because you have it in both places.

vingkan: Nevermind, I see this stands for pickup_payment, not a typo. Can you write out pickup and dropoff for these?

app/components/questions/TaxiPaymentMethod.js
```diff
@@ -0,0 +1,115 @@
...
+            <h2>Most Popular Payment Method Per Dropoff Location</h2>
+            <p>This table shows the most common payment method for taxi trips for every Dropoff location in Chicago 
+            as well as the total amount of taxi rides paid for with that method. </p>
+            <p>
+              <span> This is for the 6-month period from January 2021 - July 2021.</span>
+            </p>
+          </div>
+          <Table rows={payment_per_dropoff} cols={COMMON_DPAYMENT_COLS} />
+          
+          {errorMsg}
```
vingkan, line 112: Instead of `{errorMsg}`, can you use the new `FailureMessage` component? It will help clean up the HTML and make it more readable.

You can import it from here:

```js
import { FailureNotification } from '../../components/Notification'
```

And use it like this:

```js
<FailureNotification error={error} data={data} />
```

See `TemplateWithPieChart.js` for an example.

app/components/questions/TaxiPaymentMethod.js
```diff
@@ -0,0 +1,115 @@
...
+      payment_per_pickup,
+      payment_per_dropoff
+      ];
+  }
+  return [ [] , [] ];
+}
+
+export default function TaxiTrips(props) {
+  const { loading, error, data } = useFetch(COMMON_PAYMENT_ENDPOINT, {}, []);
+  const [ payment_per_pickup, payment_per_dropoff ] = transformData(data);
```
vingkan, line 79: Since all `transformData()` does is pull out these two lists, you could also write it like this:

```js
const { loading, error, data } = useFetch(COMMON_PAYMENT_ENDPOINT, {}, []);
const payment_per_pickup = data?.payment_per_pickup || [];
const payment_per_dropoff = data?.payment_per_dropoff || [];
```

The syntax `?.` is "optional chaining," if `data` is undefined or null, instead of throwing an error, it will return undefined.

The syntax `||` means "or" when you use it in a conditional, but it has another use:

```js
const something = ???
const output = something || "default";
```

If `something` is null or undefined, the `||` will cause the expression to evaluate to `"default"`.

What you have works, just wanted to explain some JavaScript syntax that may be new and handy to you.

vingkan (approved):
Good work Leilah! I left some optional comments, you don't have to change them in this PR, but check it out as you may find some useful for future work.

PR #166, opened 2021-08-02
Hotfix: Add dataset property to sidewalk cafe scatter metric

vingkan:
To fix the error on the production website when clicking on the metric selector.

### DIFF
--- a/app/site/metrics.js
+++ b/app/site/metrics.js
@@ -41,7 +41,6 @@ export const communityMetrics = {
   rideshare_pickups_covid: {
     name: "Rideshare Pickups Since March 2020",
     units: "trips",
-    dataset: "Rideshare",
     description: "",
     format: Formatter.numberInMillions,
     fullFormat: Formatter.numberWithCommas,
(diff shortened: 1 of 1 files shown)


fabrego524 (commented):
This seems to fix the issue with the thrown Unexpected Error. However, another error is thrown in the console, as selecting Sidewalk Cafe causes an infinite Loading Data... 

image

vingkan:
> This seems to fix the issue with the thrown Unexpected Error. However, another error is thrown in the console, as selecting Sidewalk Cafe causes an infinite `Loading Data...` on the page
> 
> image

Good catch! The error is “column area_number does not exist on table sidewalk_cafe” - Asude merged the change that adds this column, maybe it didn’t make it into the compressed db files. I reran the pipeline and that fixed it.

I also added a check to validate the metrics. That should cause the build to fail if something is wrong, so we can automatically catch this bug in pull request reviews. See the checks tab for an example failure.

fabrego524 (approved):
These changes fix both the Unexpected Error and the POST Error 500. Data loads properly and the webpage works as intentional now.

LGTM

PR #170, opened 2021-08-02
added new dataset for popular train stops, changed end date for cta train ridership data

ashklover:


pipeline/transform/popular_train_stops.py
```diff
@@ -0,0 +1,30 @@
+import sys
+sys.path.append("./")
+
+import argparse
```
vingkan, line 4: Do you want this to be a separate transform script from your existing one?

If they both do the same thing, you could change the Makefile step to use the existing transform script, but with the popular train stops inputs.

Or, if you think they will be doing different transformations in the future, it would be worth havinn them separately.

vingkan (approved):
I left one comment about the transform step, but either way looks good!

pipeline/transform/popular_train_stops.py
ashklover: It's a separate dataset which looks at only the popular train stops. I might have some different transform steps so I thought it would be a good idea to have a separate file.

PR #171, opened 2021-08-02
Fixes workflow for Github action

fabrego524:
Fixes a small syntax problem with the new Github action, should fix the workflow.


vingkan (approved):
Good fix!

PR #173, opened 2021-08-03
Added escooter to the app

talemu:
AppLaunch
AppLaunch2


vingkan (approved):
Looks good, let's deploy!

Need to look into why the conversion of meters to miles didn't work and why the map of Chicago doesn't show change in shade for the values.

PR #174, opened 2021-08-03
Upgrade Node version to 14

vingkan:
To test this change:

```
# Check out this branch
git fetch
git checkout upgrade_node
git pull
# Delete your existing dependencies
rm -rf node_modules
rm -rf app/node_modules
# Run the app
run app 9 api prod
```

If the app runs and works as expected, then this worked and you are now running the app on Node v14!

**Note:** After you stop running the app, if you type `node --version` in your terminal, you may still see an older version (probably v10), because the `run app` shortcut only changes the node version for the process it is running


fabrego524 (approved):
I ran checkout and pulled the branch, and ran the app. Everything seems to work perfectly, checked both Explore pages and a Question page, all seemed to be working as expected.

LGTM

PR #178, opened 2021-08-04
Disabilities question cleanup

MariaGarciaCS:
ADA section is now first in the page so the large table is at the bottom.

The Y axis on the bar graph no longer shows percentages, instead shows number in thousands. 
The legend on the bar graph no longer says av_trips_since, it now reads Avg Trips Since

image


vingkan (approved):
Great job, Maria!

MariaGarciaCS:
I noticed a typo and pushed the changes a few minutes after opening this pull request. The new changes include a change to html text and I deleted code that was commented out.

PR #180, opened 2021-08-05
fixed map bug

### LINKED ISSUE #175: Community area map shading fails for escooter metrics
## Description

When you view escooter metrics on the y-axis of the community scatter view, the shading of the map does not work.

## To Reproduce

- Go to the community scatter view
- Select an escooter metric as the y-axis
- View the Chicago map on the right-hand side of the screen

The entire map will be green:

Screen Shot 2021-08-03 at 5 03 45 PM

## Expected Behavior

Instead, each community area should be shaded based on its relative value:

Screen Shot 2021-08-03 at 5 02 14 PM

talemu:
Fixed #175


vingkan (approved):
Glad we fixed this!

PR #182, opened 2021-08-05
RideTrips Branch Created, RideTrips files added

wpj0110:
I was doing all the work under metrics, but I moved them all under questions and created their own files called ridetrips.py, ridetrips_test, with the class RideTrips. 

I wasn't sure how to revert back to the very original versions of metrics/rideshare.py and metrics/rideshare_test.py, and endpoints/rideshare.py so I just deleted my changes. With my main branch, I've done
git pull
git fetch
and I merged my new branch with the main branch and I found no issues. As far as I'm concerned, there should be no merge conflicts.

I should've pushed bit by bit instead of doing this all at once in hindsight. 

Still working on the RideTrips question. I'm not sure if what I've done so far is the right approach.


api/metrics/rideshare_test.1.py
```diff
@@ -2,6 +2,7 @@
 sys.path.append("../")
```
vingkan, line 2: Looks like the file `api/metrics/rideshare_test.1.py` was added by accident in one of Tabor's previous PRs and I missed that in my review. Can you delete this file?

Are the tests you added in this file identical to the other test file you added? If so, you can just delete this file, otherwise you can move the new tests over.

vingkan (changes requested):
Delete this file: `api/metrics/rideshare_test.1.py` and then I will approve!

vingkan (approved):
Good work, William!

PR #184, opened 2021-08-06
meters to miles bug is fixed and application is updated

### LINKED ISSUE #176: Convert escooter average trip distance from meters to miles
The data portal documentation for the [2019 escooter pilot data](https://data.cityofchicago.org/Transportation/E-Scooter-Trips-2019-Pilot/2kfw-zvte) says:

> Trip Distance | Trip distance in meters. (Divide by 1609 for miles.)

We tried to do this conversion the other night, thinking it would be easier for Chicago residents and Americans to understand, but we ran into a strange error where the division results were incorrect.

My theory on what happened:

- The first time we did the division, it was incorrect because we did a division by two integers, instead of two floats
- Every subsequent time we tried to change the division, we were actually viewing the original version of the output file, and not the updated version

Let's give it another try when you have a chance.

talemu:
fixed #176


pipeline/transform/escooter.py
```diff
@@ -20,17 +20,19 @@
...
 # Convert community area numbers to integers
-df["avg_trip_distance"] = df["avg_trip_distance"].astype(int)
+df["avg_trip_distance"] = df["avg_trip_distance"].astype(float)
 
 #Sort data by average trip distance
-df["avg_trip_distance"] = sorted(df["avg_trip_distance"])
+
+#transform avg trip distance from meters to miles
+df["avg_trip_distance_miles"] = df["avg_trip_distance"] / 1609.0
 print("dataframe was sorted successfully!")
```
vingkan, line 36: Looks like you deleted the line that sorts by distance. You don't really need it, so just delete the print on line 36 as well.

vingkan (changes requested):
Left a comment on your transform script.

Can you also add a screenshot to your PR of what the new values look like on the scatter plot?

talemu:
image

talemu (commented):
image

vingkan (approved):
One more change: now that the values are a thousand times smaller, how about we change format and fullFormat for those two metrics to be a number with two decimal places?

After that you are good to merge. Thanks for fixing this, Tabor!

PR #187, opened 2021-08-06
home page now display insights with functioning search

MariaGarciaCS:
Clicking on "view insight" takes user to corresponding page (scatter plot) with set x and y axis
image


vingkan (approved):
Excellent job, Maria! Looking forward to the next PR with the styles.

PR #189, opened 2021-08-08
Searchbar 3 UI

MariaGarciaCS:
Styled search bar on home page

- Should text in search bar be centered? 
- Any other changes necessary? 

image

### DIFF
--- a/app/components/InsightSearch.js
+++ b/app/components/InsightSearch.js
@@ -26,7 +26,7 @@ export default function InsightSearch(){
       <div className="searchBar center">
         <input
           type="text"
-          placeholder="What do you want to know about?"
+          placeholder="What do you want to know?"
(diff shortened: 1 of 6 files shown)


vingkan (approved):
Beautiful! This looks excellent, Maria.

Either in this PR or in another PR, can you also see how it looks on responsive mobile? I wonder if the 70% width will not work on mobile. If you can add a screenshot of the mobile view, that would be great.

MariaGarciaCS:
I had to make some changes. You were right, it was not displaying properly in mobile. This has been fixed. I added a media query with the cut off at 600px so all devices with a width smaller than that will display the button below the text instead of to the right. Additionally the search result text will be centered. 
**desktop**
image

**iphonex**
image

Issue #160, opened 2021-08-02
Modal closes when click ends outside of modal, even if input is focused.


vingkan:
## Description

Modal closes when click ends outside of modal, even if input is focused.

This is a bug in the sense that it is disruptive to the user. It's not an edge case that we anticipated, but that doesn't make it not a bug. That being said, it may be tricky to fix and is not super high priority, so we don't have to fix it now. 

## To Reproduce

- Go to scatter view or timeline view
- Click metric selector to open modal
- Type something in search bar
- Drag mouse to highlight the search bar contents
- Release mouse drag outside of modal
- Modal will close

Recording showing example:

https://www.loom.com/share/8afac30dc06f41abbaba11ae064924f5

## Expected Behavior

Modal should not close while input is focused.

## Suggested Solution

Add a state variable to keep track of whether the input is focused. This should capture when someone is typing in the input as well as when they are highlighting content in the input. If the input is focused, do not close the modal.

PR #190, opened 2021-08-10
Fix for Modal Closing when Input is Focused

fabrego524:
This should apply a fix for #160, the modal now checks if the Input is being focused or not. 

It seems this solution 

```javascript
onClick={focus ? undefined : onClose}
``` 

will work, although I'm not sure if there is a better way of doing this. Opinions or feedback on this would be great.


vingkan (approved):
Good job, Fabian! I just pulled the branch to test the fix and it works as expected. Great fix.

PR #191, opened 2021-08-10
Traffic intensity question

sperveen4:
Added pie chart to the question tap along with the year selector


api/metrics/community.py
```diff
@@ -205,6 +205,24 @@ def traffic_intensity(self, year, segment):
         rows = rows_to_dicts(cur, cur.fetchall())
         return rows
     
+    def traffic_intensity_grouping(self, year, segment):
```
vingkan, line 208: Since this is not used you can remove it.

app/components/questions/TrafficIntensity.js
```diff
@@ -0,0 +1,176 @@
...
+                  :(selectedYear === 2017) ? 'traffic_intensity_2017'
+                  :(selectedYear === 2016) ? 'traffic_intensity_2016'
+                  : 'Error');
+
+  const dataWithName = response.metrics.map(d => ({
+    // Get all fields from original record
+    ...d,
+   
+    // Add formatted data
+    traffic_intensity_formatted: parseFloat(Formatter.numberWithTwoDecimals(d[traffic_intensity_year])),
```
vingkan, line 66: You can remove the parse float. The parse float undoes the formatting.

vingkan (changes requested):
Some changes then you can merge. Great progress Shahzia!

app/components/questions/TrafficIntensity.js
Andrew-Tolentino: Yea we initially had `Formatter.numberWithTwoDecimals(d[traffic_intensity_year])` but it returned a string for `traffic_intensity_formatted`'s value which the Piechart did not like

vingkan: The formatted field should be for displaying the values outside of the chart. For the chart data, you can use the raw value `d[traffic_intensity_year]`, right? Maybe add a field:

```
{
  ...
  value: d[traffic_intensity_year],
  ...
}
```

Andrew-Tolentino: ahh I see. Yea the initial issue which led use to writing this was that there were too many significant figures for values for our dataset which led us to use the Formatter. Do you think it would be better to `ROUND` on the backend within the SQL query? @vingkan

vingkan: Oh, I see. What you did with the round should work, but there could still be situations where you get trailing decimals when the response is received by the frontend.

Ultimately, what we'd want to do is specify to the Recharts visualization how to format the data it displays. I don't know off the top of my head what it is, but if you look at the documentation or examples online you can find out how to pass a formatter function to it.

But that's not super important here, so let's merge the solution you and Shahzia did that is working and move on to adding Urdu support in the next PR.

Andrew-Tolentino (approved):

vingkan (approved):
LGTM

PR #192, opened 2021-08-12
Rideshare trips question added

wpj0110:
Still missing visuals, will try to do this


vingkan (commented):
Can you add a screenshot of the page to your PR description?

I will pull and check out the changes on my Cloud9 instance.

wpj0110:
temp7

wpj0110:
temp7

api/questions/ridetrips.py
```diff
@@ -118,4 +118,35 @@ def get_total_trips_by_dropoff_specific_area_and_year(self, year, pickup_area):
...
+        Returns the rows, but with a new column that contains the area names
+        """
+        ls = self.community_areas()
+        
+        for area in ls:
+            y = area['area_number'] 
+            #x = rows[y-1]['dropoff_community_area']
+            x = rows[y-1]['dropoff_community_area']
+            if x == y:
+                rows[y-1]['area_name'] = area['name']
```
vingkan, line 150: If x is not equal to y (in the case that `rows` is not sorted, you will miss some area names.

Two options:
1. Revise this Python method
2. Add area name in the SQL by doing a left join

I think option 1 would work fine. If you do option 2, then you would also need to add some fake community areas to your unit test, which might be extra work for little benefit.

Here's the pseudocode I would recommend for option 1:
- Convert the list of community areas into a dictionary where the key is the area number and the value is the object with the name, number, and part.
- Loop over all the values in rows and look up the dropoff area in the dictionary
- Add the name, number, and part info to the output (or just the name if that's all you need)

api/endpoints/question.py
```diff
@@ -3,6 +3,7 @@
 from api.questions.pooled_trips import PooledTripMetrics
 from api.questions.taxitrips import TaxiTripQuestions
 from api.questions.sidewalk_search import SidewalkCafePermitSearch
+from api.questions.ridetrips import RideTrips
```
vingkan, line 6: Can you rename this class to be something like `OHareRideshareQuestion`?

api/endpoints/question.py
```diff
@@ -65,5 +67,23 @@ def question_sidewalk_search():
             return jsonify({ "results": permits })
         return jsonify({ "results": [] })
 
+    @app.route("/question/ridetrips")
```
vingkan, line 70: Similarly, can we rename this endpoint to something like `/question/ohare/rideshare`?

app/components/questions/RideTripsOhare.js
```diff
@@ -0,0 +1,130 @@
...
+import { FailureNotification } from '../../components/Notification'
+import {
+  Formatter,
+  calculatePercentChange,
+} from '../../site/metrics'
+
+const RIDETRIPS_ENDPOINT = `${process.env.NEXT_PUBLIC_API}/question/ridetrips`;
+  
+  const MAX_PCT_CHANGE = -0.76;
+  const RIDETRIPS_DROPOFF_ALL = [
```
vingkan, line 14: Good job setting up the table config! Can you move the indentation of lines 13, 14, and 43 back one indent?

app/components/questions/RideTripsOhare.js
```diff
@@ -0,0 +1,130 @@
...
+  
+  const dropoffs_2020 = since.map(({
+    total_trips: total_trips_2020,
+
+  }) => ({
+    total_trips_2020,
+
+  }));
+  
+  //adding 2020 trips within the same object as the 2019 trips
```
vingkan, line 66: This also assumes that the results are in the same order by area number. Would you rather write this logic on the backend in Python or on the frontend in JavaScript?

app/components/questions/RideTripsOhare.js
```diff
@@ -0,0 +1,130 @@
+import { useState, useEffect } from 'react'
+import useFetch from 'use-http'
+import { Table } from '../../components/Table'
+import { Notification } from '../../components/Notification'
+import { FailureNotification } from '../../components/Notification'
```
vingkan, line 5: You can use `FailureNotification` instead of `Notification` and remove `Notification` entirely.

app/site/questions.js
```diff
@@ -12,6 +12,7 @@ import TaxiMostCommonDropoff from '../components/questions/TaxiMostCommonDropoff
 import TaxiPaymentMethod from '../components/questions/TaxiPaymentMethod'
 import EscooterCity from '../components/questions/EscooterCity'
 import CovidCasesPerRideshare from '../components/questions/CovidCasesPerRideshare'
+import RideTrips from '../components/questions/RideTripsOhare'
```
vingkan, line 15: Can we rename this to something like `OHareRideshares`? Both the component you export and the file should have the same name.

app/site/questions.js
```diff
@@ -119,6 +121,12 @@ export const questionsParams = {
     component: "traffic-intensity-across-the-city",
     description: "What is the vehical density in Chicago for a given year?",
   },
+  "ridetrips-ohare": {
+    title: "Ride Trips from O'hare",
```
vingkan, line 125: Is it O'Hare or O'hare?

app/components/questions/RideTripsOhare.js
```diff
@@ -0,0 +1,130 @@
...
+        <h2>Most Common Dropoff Locations</h2>
+        <p>This table shows the dropoff trips for each locations from O'hare for the year {props.year}.</p>
+        <Table rows={props.rRow} cols={props.cColumn} />
+        {errorMsg}
+      </div>
+    </div>
+    );
+}
+  
+  export default function RideTrips(props) {
```
vingkan, line 110: Move the indentation of line 110 back to the start of the line.

app/components/questions/RideTripsOhare.js
```diff
@@ -0,0 +1,130 @@
...
+      <p>Failed to get data from server. Please reload.</p>
+    </Notification>
+  ) : null;
+  
+  return(
+    <div className="center medium-width">
+    <TableDropoff rRow={ohare_dropoff} cColumn={RIDETRIPS_DROPOFF_ALL}/>
+    {errorMsg}
+    </div>
+    );
```
vingkan, line 129: Move the indentation of line 129 back one level, to match line 124.

app/components/questions/RideTripsOhare.js
```diff
@@ -0,0 +1,130 @@
...
+    <Notification classes={["Bottom", "Wide", "Failure"]} visible={true}>
+      <p>Failed to get data from server. Please reload.</p>
+    </Notification>
+  ) : null;
+  
+  return (
+    <div className="DropoffTrips">
+      <div className="center medium-width">
+        <h2>Most Common Dropoff Locations</h2>
+        <p>This table shows the dropoff trips for each locations from O'hare for the year {props.year}.</p>
```
vingkan, line 102: It would be good to explain that these are rideshares to the O'Hare area, not all of them may be to/from the airport.

As well as more background on why we would even want to know about this.

vingkan (changes requested):
Excellent start, William. I left some comments, in descending order of importance:

- Most important: there is code on both the frontend and backend that assumes that data will be in order by community area. That may be true, but I would rather we not assume that.
- Also important: Add more detail to the data question to explain why the question is important and what the data means.
- Less important: Updating the names of the classes, as RideTrips is too vague.
- Least important: Fixing indentation.

api/questions/ridetrips.py
wpj0110: With option 1:
Will this suffice?
`    def metrics_by_area(self,rows):
        """
        Returns the rows, but with a new column that contains the area names
        """
        res = {}
        
        for area in self.community_areas():
            res[area["area_number"]] = area
        for row in rows:
            if row['dropoff_community_area'] in res.keys():
                row['area_name'] = res.get(row['dropoff_community_area'])['name']
                row['part'] = res.get(row['dropoff_community_area'])['part']
        return rows`

I have tested this and it still works just fine for the frontend

wpj0110:
Screenshot 2021-08-12 2 46 27 PM
New look of the page. Including the changes on the python method

app/site/questions.js
```diff
@@ -12,6 +12,7 @@ import TaxiMostCommonDropoff from '../components/questions/TaxiMostCommonDropoff
 import TaxiPaymentMethod from '../components/questions/TaxiPaymentMethod'
 import EscooterCity from '../components/questions/EscooterCity'
 import CovidCasesPerRideshare from '../components/questions/CovidCasesPerRideshare'
+import RideShare from '../components/questions/RideshareTripsOHare'
```
vingkan, line 15: Change the name of this to `RideshareTripsOHare`, both on line 15 and line 26.

app/components/questions/RideshareTripsOHare.js
```diff
@@ -0,0 +1,135 @@
...
+        <h2>Dropoff Trips From The O'Hare Area to Places in Chicago</h2>
+        <p>This table shows the dropoff trips from the O'Hare Area to different locations in Chicago for 2019 and 2020.</p>
+        <Table rows={props.rRow} cols={props.cColumn} />
+        {errorMsg}
+      </div>
+    </div>
+  );
+}
+  
+export default function RIDESHARE(props) {
```
vingkan, line 115: Change the name of this from RIDESHARE to `RideshareTripsOHare` to match what will be imported.

wpj0110: ok

app/components/questions/RideshareTripsOHare.js
```diff
@@ -0,0 +1,135 @@
...
+  const dropoffs_2020 = since.map(({
+    total_trips: total_trips_2020,
+
+  }) => ({
+    total_trips_2020,
+
+  }));
+  
+  //adding 2020 trips within the same object as the 2019 trips
+  for (var i = 0; i < dropoffs.length ; i++){
```
vingkan, line 67: We still have the issue here of assuming the data is (a) sorted and (b) has all the community areas with no nulls. How do you want to handle it?

Do you want to solve it on the frontend or on the backend?

If you go with a frontend solution:
- Make a dictionary and use area number as a key
- Then convert the dictionary to a list
- Similar to what you did when fixed the backend code

If you go with a backend solution:
- You could rewrite the SQL query to return data in the format you want
- Or you could use Python to merge the results, again similar to what you did to fix the other problem

vingkan (changes requested):
Left some comments! The big one to fix is in the transform data function.

wpj0110:
Added part, and I also think I solved the issue of "assuming" they match.
Screenshot 2021-08-12 7 07 25 PM

app/components/questions/RideshareTripsOHare.js
```diff
@@ -0,0 +1,132 @@
...
+
+function addTotalTripsYear(before,since) {
+
+  var ds = {};
+  for(var key in before) {
+    ds[before[key]["dropoff_community_area"]] = before[key];
+  }
+  for(var key in since) {
+    var sinceAreaNum = since[key]["dropoff_community_area"];
+    if (ds[sinceAreaNum]["area_name"] == since[key]["area_name"] && ds[sinceAreaNum]["dropoff_community_area"] == since[key]["dropoff_community_area"]){
```
vingkan, line 57: I know you were being extra cautious, but this check is not needed.

app/components/questions/RideshareTripsOHare.js
```diff
@@ -0,0 +1,132 @@
...
+    rgb: "209, 31, 0",
+    alpha: (v) => (v / MAX_PCT_CHANGE),
+  },
+];
+
+function addTotalTripsYear(before,since) {
+
+  var ds = {};
+  for(var key in before) {
+    ds[before[key]["dropoff_community_area"]] = before[key];
```
vingkan, line 53: If you want, you could do the same thing you do below and rename this "total_trips_before" instead of just "total_trips"

app/components/questions/RideshareTripsOHare.js
```diff
@@ -0,0 +1,132 @@
...
+      ds[sinceAreaNum]["total_trips_since"] = since[key]["total_trips"];
+      ds[sinceAreaNum]["pct_change"] = calculatePercentChange(ds[sinceAreaNum]["total_trips"],ds[sinceAreaNum]["total_trips_since"]);
+    } 
+    else {
+      console.log("Something wrong. Mismatching area names and area numbers from both datas");
+    }
+  }
+  
+  const fin = Object.values(ds); //turns key:value of the original object into elements of an array
+  console.log("fin is ",fin)
```
vingkan, line 69: Remove the console.log.

vingkan (changes requested):
Awesome, great job fixing the frontend issue! Clean up the remaining comments I opened and then I will approve.

vingkan (approved):
Great work, looks good to merge!

PR #193, opened 2021-08-12
extracted, transformed, and loaded covid data

talemu:


divyashie (approved):
LGTM! 👍🏽

divyashie:
Not sure why some of the checks are failing here, will have to take a closer look. Maybe hold off before you merge this @talemu

pipeline/extract/Covid_cases_deaths_data.sql
```diff
@@ -0,0 +1,25 @@
...
+    sum(deaths_0_17_yrs),
+    sum(deaths_18_29_yrs),
+    sum(deaths_30_39_yrs),
+    sum(deaths_40_49_yrs),
+    sum(deaths_50_59_yrs),
+    sum(deaths_60_69_yrs),
+    sum(deaths_70_79_yrs),
+    sum(deaths_80_yrs),
+    Min(lab_report_date) as week_start,
+    Max(lab_report_date) as week_end
```
vingkan, line 21: Can you add a where clause to restrict to only rows less than or equal to a certain lab report date? The lab report date for that where clause should be whatever is the current most recent report date for this dataset, which you can check on the data portal: https://data.cityofchicago.org/Health-Human-Services/COVID-19-Daily-Cases-Deaths-and-Hospitalizations/naz8-j4nc

The city updates this dataset very frequently, so we want to control when we get updates. For more information on why this is important and can cause problems, see #5

vingkan (changes requested):
Good job doing the ETL! I left comments on what needs to be improved.

vingkan:
> Not sure why some of the checks are failing here, will have to take a closer look. Maybe hold off before you merge this @talemu

Test failures:

```
self = <pipeline.utils.testing.DatabaseTestHelper object at 0x7f1ccc943110>
table = 'escooter', expected_count = 411

    def assert_table_count(self, table, expected_count):
        """
        Helper method to check the size of a table.
        Note: Unsafely injects table name into query.
        """
        cur = self.con.cursor()
        cur.execute("""
            SELECT count(1) AS count
            FROM {table};
        """.format(table=table))
        actual_count = cur.fetchone()[0]
        msg = f"Table `{table}` should have {expected_count} rows, not {actual_count} rows."
>       assert actual_count == expected_count, msg
E       AssertionError: Table `escooter` should have 411 rows, not 0 rows.

actual_count = 0
cur        = <sqlite3.Cursor object at 0x7f1ccc9c4420>
expected_count = 411
msg        = 'Table `escooter` should have 411 rows, not 0 rows.'
self       = <pipeline.utils.testing.DatabaseTestHelper object at 0x7f1ccc943110>
table      = 'escooter'

pipeline/utils/testing.py:25: AssertionError
```

Make sure you have the latest from main and rerun the pipeline.

api/metrics/Covid_cdd_metric.py
```diff
@@ -8,23 +8,14 @@ class CommunityMetrics:
     def __init__(self, con):
         self.con = con
 
-    def cases_by_age_given_month(self):
+    def cases_by_age_given_week(self, date):
```
vingkan, line 11: For the timeline chart, you don't need a metric that takes in date as a parameter, you want to return all the dates.

vingkan (approved):
Excellent work, Tabor. Fix the issue with the tests not passing and then you can merge.

The test was failing because the escooter table was missing rows. To solve that, I would recommend merging `main` into your branch, resolving any code merge conflicts, and then using `run pipeline-quick` to get the updated compressed database files. That way the tests will pass both locally and on GitHub.

talemu:
> Excellent work, Tabor. Fix the issue with the tests not passing and then you can merge.
> 
> The test was failing because the escooter table was missing rows. To solve that, I would recommend merging `main` into your branch, resolving any code merge conflicts, and then using `run pipeline-quick` to get the updated compressed database files. That way the tests will pass both locally and on GitHub.

I merged from main and ran pipeline-quick but some tests still do not pass.

vingkan (approved):
Good work on the fixes, Tabor!

vingkan:
> I merged from main and ran pipeline-quick but some tests still do not pass.

It turns out that new data got published to the Chicago Health Atlas! The belonging data got a new year: 2020 (skipping 2019 lol) and now has 237 rows instead of 200.

We don't have a solution for limiting new data updates from the Atlas like we do for the Data Portal (see #5), so that might be useful to avoid this kind of surprise in the future.

Thanks Tabor for updating the tests to reflect the latest data.

PR #194, opened 2021-08-13
made changes to prevent SQL injection

Jmacias2019:
LGTM

### DIFF
--- a/api/questions/sidewalk_search.py
+++ b/api/questions/sidewalk_search.py
@@ -9,30 +9,21 @@ class SidewalkCafePermitSearch:
         self.con = con
         
     def search_permits(self, search):
+        search = search[:50] # Limit to 50 characters
+        search = search.replace("_", "").replace("%", "").replace(";", "").replace("--", "") # Line 13 will remove these special characters 
         """
         Returns permits for restaurants that match the search query.
         """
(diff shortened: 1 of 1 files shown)


api/questions/sidewalk_search.py
```diff
@@ -9,30 +9,21 @@ def __init__(self, con):
         self.con = con
         
     def search_permits(self, search):
+        search = search[:50] # Limit to 50 characters
+        search = search.replace("_", "").replace("%", "").replace(";", "").replace("--", "") # Line 13 will remove these special characters
```
vingkan, line 13: Can you add one more replace to replace single quotes `"'"` with an empty string `""`? That way, queries with an apostrophe won't cause an error. For example, `popeye's` will become `popeyes`.

Screen Shot 2021-08-13 at 6 35 37 PM

vingkan (changes requested):
Thank you for fixing this vulnerability, Jackie! I tested it on my Cloud9 and I can no longer drop tables or inject other SQL code.

It was a good thing that you removed underscores, because that thwarted an attack I wanted to try that would let me read other data tables.

If you remove single quotes, that should close off the ability of an attacker to end the string and also fix the error that it causes.

PR #196, opened 2021-08-17
Traffic intensity question urdu translation

sperveen4:
Made some minor changes


vingkan (approved):
Thanks for fixing these, Shahzia!

PR #198, opened 2021-08-18
Minor Searchbar Cosmetic Fix

fabrego524:
Provides a very small fix to the CSS Styling for the modal's search bar, small bits of text will no longer be visible above the search.

Example of what the search bar looked like before:

image

And with the fix applied:

image

The issue still persists on some mobile devices, but it's difficult to account for all screen sizes. This was mostly a small nitpick I decided to fix.


vingkan (approved):
Glad you fixed this! This is one of those UI bugs that you just can't get off your mind once you see it haha

